PR #1, opened 2014-03-17
Ch 317: Allow access to explore-rate

eshtadc:
- Enable access to adjust explore rate
- Adjustments to control rate UI
- Convert control rate to initial intent (no more inversion)
- Set control rate default to .1
- Update tests

### DIFF
--- a/plugins/agent_types/AcquiaLiftAgent.inc
+++ b/plugins/agent_types/AcquiaLiftAgent.inc
@@ -183,7 +183,45 @@ class AcquiaLiftAgent extends PersonalizeAgentBase implements AcquiaLiftAgentInt
     $form['#attached'] = array(
       'css' => array(
         drupal_get_path('module', 'acquia_lift') . '/css/personalize_acquia_lift_admin.css',
-      )
+      ),
+      'js' => array(
+        drupal_get_path('module', 'acquia_lift') . '/js/acquia_lift.agent.admin.js',
+      ),
+    );
+    $form['control'] = array(
+      '#type' => 'fieldset',
+      '#collapsible' => TRUE,
+      '#collapsed' => TRUE,
+      '#attributes' => array('class' => array(
+        'horizontal',
+        'acquia-lift-more-information',
+      )),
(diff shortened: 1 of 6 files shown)


plugins/agent_types/AcquiaLiftAgent.inc
```diff
@@ -223,12 +299,23 @@ class AcquiaLiftAgent extends PersonalizeAgentBase implements AcquiaLiftAgentInt
    */
   public function postSave($old_data) {
     $items = array();
-    $acquia_lift_control_rate = 0;
-    if (isset($this->data['control_rate'])) {
+    $acquia_lift_control_rate = .1;
+    $acquia_lift_explore_rate = .2;
+    dpm($this->data, 'postSave data');
```
katbailey, line 304: Oops :)

plugins/agent_types/AcquiaLiftAgent.inc
```diff
@@ -316,7 +404,7 @@ class AcquiaLiftAgent extends PersonalizeAgentBase implements AcquiaLiftAgentInt
     );
     // We only show the lift over default if we are using a non-zero percentage of
     // users as a control group.
-    $show_default = isset($this->data['control_rate']) && (int) $this->data['control_rate'] > 0;
+    $show_default = isset($this->data['control']['control_rate']) && (int) $this->data['control']['control_rate'] > 0;
```
katbailey, line 407: Just curious why we're now nesting the control rate one level deeper inside the data array?

plugins/agent_types/AcquiaLiftAgent.inc
eshtadc, line 304: Darn it.

plugins/agent_types/AcquiaLiftAgent.inc
eshtadc, line 407: It had to do with the nested structure necessitated for the collapsing stuff.  I don't like it nested deeper though.  How can I keep the data structure the same?  It seems it copies directly from the form.

PR #5, opened 2014-03-27
Ch 327: Personalize campaign design and usability changes

eshtadc:


js/acquia_lift.agent.admin.js
```diff
@@ -19,41 +25,47 @@
...
+
+        // Set initial "rest" percentage value.
+        updateRest();
       })
     }
   };
 
+  /**
+   * Adds collapsible functionality into a container.
+   * @type {{attach: attach}}
```
katbailey: Probably just ignorance on my part, but what does this line mean?

eshtadc: Hmm - looks like my IDE decided to JSDoc things properly, but I didn't fill it out.  It's the beginning of JSDoc object literal documentation.  Let me see what the Drupal standards say on this, but given that we're not documenting that way anywhere else, I should take it out.

eshtadc: Hmmm seems like Drupal JS coding standards don't care a thing about this so I'll remove it.

PR #8, opened 2014-04-02
Support the new verification method in personalize module

katbailey:


plugins/agent_types/AcquiaLiftAgent.inc
```diff
@@ -623,24 +623,25 @@ class AcquiaLiftAgent extends PersonalizeAgentBase implements AcquiaLiftAgentInt
...
     if (empty($goals) || empty($acquia_lift_goals)) {
       // Acquia Lift agents need goals.
-      return FALSE;
+      $errors[] = t('No goals have been set up for this agent');
     }
 
     foreach($goals as $goal) {
       if (!in_array($goal['action'], $acquia_lift_goals)) {
-        return FALSE;
+        $errors[] = t('Goal @goal has not been sync\'d to the Acquia Lift agent', array('@goal' => $goal['action']));
```
eshtadc, line 644: In other error messages I've seen where we point people towards cron to update their out of sync agent.  Would that be useful here (and below) or just out of place?

PR #10, opened 2014-04-04
Issue #2233759: Make it possible to reset all data for an agent from the...

katbailey:
... campaign page

### DIFF
--- a/acquia_lift.module
+++ b/acquia_lift.module
@@ -967,6 +967,58 @@ function acquia_lift_navbar() {
 function acquia_lift_form_personalize_agent_form_alter(&$form, &$form_state) {
   module_load_include('inc', 'acquia_lift', 'acquia_lift.ui');
   acquia_lift_chosenify_element($form['agent_form']['agent_fieldset']['visitor_context']);
+  // If we're editing an existing agent, add a "Reset data" button next to
+  // the Pause/Resume button.
+  if (isset($form['agent_form']['agent_fieldset']['header']['toggle_form'])) {
+    $subform = array(
+      '#prefix' => '<div id="personalize-acquia-lift-reset-form">',
+      '#suffix' => '</div>'
+    );
+    $subform['actions']['reset'] = array(
+      '#prefix' => '<div id="personalize-acquia-lift-reset">',
+      '#suffix' => '</div>',
+      '#type' => 'submit',
+      '#name' => 'reset_submit',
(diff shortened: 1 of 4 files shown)


acquia_lift.module
```diff
@@ -967,6 +967,58 @@ function acquia_lift_navbar() {
...
+      '#prefix' => '<div id="personalize-acquia-lift-reset">',
+      '#suffix' => '</div>',
+      '#type' => 'submit',
+      '#name' => 'reset_submit',
+      '#value' => t('Reset data'),
+      '#attributes' => array(
+        'class' => array('action-item-primary-active'),
+      ),
+      '#ajax' => array(
+        'callback' => 'personalize_acquiq_lift_ajax_callback',
```
eshtadc, line 987: acquiq (although you are consistent - the callback is acquiq too)

includes/acquia_lift.classes.inc
```diff
@@ -415,6 +415,25 @@ class AcquiaLiftAPI implements PersonalizeLoggerAwareInterface {
...
+    $vars = array('agent' => $agent_name);
+    $success_msg = 'The data for Acquia Lift campaign {agent} was reset';
+    $fail_msg = 'Could not reset data for Acquia Lift campaign {agent}';
+    if ($response->code == 200) {
+      $this->logger()->log(PersonalizeLogLevel::INFO, $success_msg, $vars);
+    }
+    else {
+      $this->logger()->log(PersonalizeLogLevel::ERROR, $fail_msg, $vars);
+    }
+  }
```
eshtadc, line 434: This is probably an issue on its own instead, but here we know if there is an error or not, but we don't return an error code.  Instead in the calling function we check for an exception being thrown, but couldn't the error just be that we get an invalid response code?  Seems like the user could get a message that the data was reset when it wasn't.

acquia_lift.module
katbailey, line 987: Copy/paste has its benefits: my typos still _work_ :) Nice catch

includes/acquia_lift.classes.inc
katbailey, line 434: Ooh, yeah that is dumb. Most of the PUT/DELETE requests are made asynchronously by the queue runner so we don't bother throwing an exception. It does make sense to throw it here though. The queue processor catches any exception thrown so maybe I should make them all throw an exception on failure.

PR #11, opened 2014-04-08
Ch 344: Split reports into four areas and add graphical indicators

eshtadc:


acquia_lift.ui.inc
```diff
@@ -69,15 +69,21 @@ function acquia_lift_build_page(&$page) {
  *
  * @param array $element
  *   An array representing a multi-select form element.
+ * @param string $size
+ *   The size of the chosen element to create: one of 'full', 'half'.
  */
-function acquia_lift_chosenify_element(&$element) {
+function acquia_lift_chosenify_element(&$element, $size = 'full') {
```
katbailey: Not a big deal, but it feels like this new param should just be a boolean, i.e. $full = TRUE.

acquia_lift.ui.inc
```diff
@@ -69,15 +69,21 @@ function acquia_lift_build_page(&$page) {
...
   $chosen_path = libraries_get_path('chosen');
+  $class = array('acquia-lift-chosen-select');
+  if ($size === 'half') {
+    $class[] = 'acquia-lift-chosen-select-half';
+  }
```
katbailey: Or maybe just pass in an array of classes to add?

plugins/agent_types/AcquiaLiftAgent.inc
```diff
@@ -84,6 +131,30 @@ interface AcquiaLiftAgentInterface {
 class AcquiaLiftAgent extends PersonalizeAgentBase implements AcquiaLiftAgentInterface, PersonalizeExplicitTargetingInterface, PersonalizeAgentReportInterface {
 
   /**
+   * The threshold to use above which lift percentage will be positively noted.
+   * @var int
+   */
+  static public $lift_threshold = 0;
```
katbailey: Seems like these should really be class constants rather than public variables. How can different values be specified from the UI?

plugins/agent_types/AcquiaLiftAgent.inc
```diff
@@ -42,30 +42,77 @@ interface AcquiaLiftAgentInterface {
...
+   * @param $date_end
+   *   The end date of hte report.
+   * @param $decision_point
+   *   The decision point to use to filter report data.
+   * @param array @selected
+   *   The selected context codes.
+   * @return array
+   *   A render array representing the variation set report.
+   */
+  public function buildReportContextSelection($date_start, $date_end, $decision_point);
```
katbailey, line 115: Not a comment that needs to be acted on for this sprint, but seeing all the changes to the AgentLiftAgent, it's clear to me that we really need an AcquiaLiftReport class that takes an agent as a parameter. This is a refactoring we can do next sprint.

plugins/agent_types/AcquiaLiftAgent.inc
katbailey: Actually, if we're going to refactor this stuff anyway then there's no need to make any change here yet.

eshtadc: Too late ;-)

acquia_lift.ui.inc
eshtadc: It was to allow for the centralization of css classes for sizes, but i like the array of classes idea better since there is already a need for that.

PR #13, opened 2014-04-11
Ch 388: Refactor reporting into its own class.

eshtadc:
Moves reporting functionality into it's own class.
Updates lift report page to use new class.

### DIFF
--- a/acquia_lift.admin.inc
+++ b/acquia_lift.admin.inc
@@ -299,6 +299,30 @@ function acquia_lift_report($form, &$form_state, $agent_data, $option_set) {
       )
     );
   }
+  // Determine possible goal value for entire campaign
+  $goals = personalize_goal_load_by_conditions(array('agent' => $agent_data->machine_name));
+  $total_goal_value = 0;
+  foreach($goals as $goal) {
+    $total_goal_value += $goal['value'];
+  }
+  // Add the final goals value total row to the report.
+  $reports['experiment']['content']['#rows'][] = array(
+    'data' => array(
+      '',
+      '',
+      '',
+      array (
+        'data' => t('Out of a possible @goal_value', array('@goal_value' => $total_goal_value)),
+        'class' => array('acquia-lift-experiment-goal-total'),
+      ),
+      '',
+      '',
+      '',
+    ),
+    'class' => array('acquia-lift-experiment-footer-row'),
+    'no_striping' => TRUE,
+  );
+  // Render the report.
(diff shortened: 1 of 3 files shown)


plugins/agent_types/AcquiaLiftAgent.inc
```diff
@@ -455,116 +372,515 @@ class AcquiaLiftAgent extends PersonalizeAgentBase implements PersonalizeAgentGo
...
+   * @param $date_to
+   *   The ending date for report data in Y-m-d format, or null for today.
+   * @return array
+   *   The reporting data for the date range as an array with keys for
+   *   - status: The general agent status report data
+   *   - confidence: The confidence report data
+   *   - targeting: The targeting report data
+   */
+  protected function loadReportData($decision_name = NULL, $date_from = NULL, $date_to = NULL) {
+    $report_data = &drupal_static(__FUNCTION__);
```
katbailey: No need for drupal_static here, just store it in an object property.

acquia_lift.admin.inc
```diff
@@ -215,7 +224,7 @@ function acquia_lift_report($form, &$form_state, $agent_data, $option_set) {
     $decision_point = $form_state['values']['decision_point'];
   }
   else {
-    $decision_point = !empty($option_set) ? $option_set->decision_name : key($decisions);
+    $decision_point = !empty($option_set) ? personalize_get_decision_name_for_option_set($option_set) : key($decisions);
```
katbailey, line 227: I think it might be safe to grab the decision name directly off the option set (I'm pretty sure it gets set during option_set_load). The reason I mention it is that this call to personalize_get_decision_name_for_option_set() and a call to personalize_get_option_label_for_decision_and_choice() on line 1341 look like they're the only things standing in the way of this class being completely unit testable. It would be awesome to be able to write blazingly fast tests for these methods just by passing in a dummy agent and a dummy liftAPI class.

eshtadc, line 227: Actually - this specific line is not in the reports class so it shouldn't matter so much.  I've made some changes to remove the icky personalize_\* functions from the report class so that it is encapsulated.

eshtadc, line 227: To do this I moved the overall goal count line out of the reporting class (which does feel a bit odd).  I'm on the fence about doing this or just passing in the number of goals when creating the report (although I don't believe this is required for all agents so that feels icky too).  What do you think?

PR #16, opened 2014-04-23
Ch 393: Reports update

eshtadc:
- Implement predicted goal value
- Bug fixes
- Nomenclature changes
- Refactoring in order to prepare for client-side model handling
- CSS cleanup in report headings


acquia_lift.admin.inc
```diff
@@ -357,39 +389,55 @@ function acquia_lift_report($form, &$form_state, $agent_data, $option_set) {
...
+        '#type' => 'container',
+        '#attributes' => array(
+          'class' => array('acquia-lift-report-header-summary'),
+        ),
       ),
     ),
-    '#theme_wrappers' => array('container'),
     '#attributes' => array(
-      'id' => 'acquia-lift-stability-report',
+      'id' => 'acquia-lift-stability-reort',
```
katbailey: interesting :P

plugins/agent_types/AcquiaLiftAgent.inc
```diff
@@ -849,11 +870,126 @@ class AcquiaLiftReport implements PersonalizeAgentReportInterface {
...
+    if (!is_array($report_data['status']) || !is_array($report_data['confidence']) || !is_array($report_data['targeting']) || !is_array($report_data['potential_context'])) {
+      return;
+    }
+    $agent_data = $this->agent->getData();
+    $isAdaptive = $agent_data['decision_style'] === 'adaptive';
+
+    // Determine overall confidence based on confidence in choices.
+    $total_confident = 0;
+    foreach($report_data['confidence']['features'][self::NO_FEATURES] as $choice) {
+      if ($choice['significant'] == TRUE) {
```
katbailey: Unless you meant to put `=== TRUE` here, it should just be `if ($choice['significant'])`. Small thing, but it looks funny.

plugins/agent_types/AcquiaLiftAgent.inc
```diff
@@ -697,6 +697,21 @@ class AcquiaLiftReport implements PersonalizeAgentReportInterface {
...
+
+  /**
+   * Used to indicate that a reporting value is negative.
+   */
+  const NEGATIVE = FALSE;
+
+  /**
+   * Used to indicate that a reporting value is not positive or negative.
+   */
+  const NEUTRAL = '';
```
katbailey: I find this quite a confusing use of constants. Could we maybe just define them as 'positive', 'negative' and 'neutral' instead of using booleans for two-thirds of the 3 values we need?

plugins/agent_types/AcquiaLiftAgent.inc
```diff
@@ -900,44 +1027,44 @@ class AcquiaLiftReport implements PersonalizeAgentReportInterface {
...
       '#attributes' => array(
         'id' => 'acquia-lift-overview-lift',
       ),
-      '#title' => $this->formatReportPercentage($lift),
+      '#title' => $overview_report['total_lift'],
       '#description' => t('Predicted lift/control'),
       '#attributes' => array(
         'class' => array(
-          $lift > self::LIFT_THRESHHOLD && $total_confident > 0 ? 'acquia-lift-report-positive' : 'acquia-lift-report-negative',
+          $overview_report['total_lift_positive'] === self::POSITIVE ? 'acquia-lift-report-positive' : 'acquia-lift-report-negative',
```
katbailey: Doesn't this ignore the possibility of neutral lift?

plugins/agent_types/AcquiaLiftAgent.inc
```diff
@@ -849,11 +870,126 @@ class AcquiaLiftReport implements PersonalizeAgentReportInterface {
...
+      $targeting_data = $report_data['targeting'][$feature_string];
+
+      // Create a hash of choice numbers for stability report.
+      // Don't show system-defined features.
+      if ($targeting_data['system'] === TRUE) {
+        continue;
+      }
+      foreach($feature as $choice_id => $choice) {
+        if ($choice['control']) {
+          $lift_default_positive = self::NEUTRAL;
```
katbailey: This seems to be the only place the neutral value is assigned and it is never checked for anywhere. So maybe we do just need to treat this as binary? All of the checks seem to be binary, i.e. if not positive then negative. If that's the case then maybe we don't need the constants at all and we can just use booleans? I'm probably missing something but figured it was worth asking...

plugins/agent_types/AcquiaLiftAgent.inc
katbailey: Er, see above comment re possibly not needing the neutral concept at all...

acquia_lift.admin.inc
eshtadc: Lol - looks all the more interesting when its a combination of commits that ends up with this as the final product.  <sigh>

plugins/agent_types/AcquiaLiftAgent.inc
eshtadc: Well - it is checked basically in absence around line 1240...but I think I'm going to change it to make these straight binary (which was how they were originally intended anyway).

eshtadc:
Merged.

PR #17, opened 2014-04-28
Ch 415

katbailey:


acquia_lift.module
```diff
@@ -1028,18 +1028,6 @@ function acquia_lift_form_personalize_agent_option_sets_form_alter(&$form, &$for
       'css' => array(drupal_get_path('module', 'acquia_lift') . '/css/acquia_lift.admin.css'),
     );
     module_load_include('inc', 'acquia_lift', 'acquia_lift.ui');
```
eshtadc: Is there any reason to load the ui file here anymore?  Looks like it was only included to call the chosenify function.

PR #18, opened 2014-04-29
Ch 402: Stylize the Acquia Lift controls when navbar not in use

eshtadc:


acquia_lift.install
```diff
@@ -16,7 +16,134 @@ function acquia_lift_install() {
     'description' => 'The <em>Acquia Lift personalize controls</em> menu contains actions to create and manage personalization campaigns.',
   );
   menu_save($menu);
+  _acquia_build_menu_link($menu_name);
+}
+
+/**
+ * Implements hook_enable().
+ */
+function acquia_lift_enable() {
```
eshtadc, line 25: NOTE:  The only change in here is moving the menu definition to a helper class - but since it is placed before the install_hook_N functions, Github is making it look like all the other functions changed.....
Changes in this file are to acquia_lift_install, _acquia_build_menu_link (new), and acquia_lift_update_7004.

acquia_lift.install
```diff
@@ -16,7 +16,134 @@ function acquia_lift_install() {
     'description' => 'The <em>Acquia Lift personalize controls</em> menu contains actions to create and manage personalization campaigns.',
   );
   menu_save($menu);
+  _acquia_build_menu_link($menu_name);
```
katbailey, line 19: Looks like this should be _acquia_lift_build_menu?

acquia_lift.install
```diff
@@ -136,7 +263,7 @@ function acquia_lift_install() {
     'menu_name' => $menu_name,
     'options' => array(
       'attributes' => array(
-        'classes' => array('acquia-lift-results-list', 'visitor-actions-ui-ignore'),
```
katbailey, line 139: wow - were these all totally broken before?

eshtadc, line 139: Yup.... they were added to the link, but as "classes" instead of "class" so I couldn't hook into them as selectors.

acquia_lift.install
eshtadc, line 19: Ack

PR #19, opened 2014-05-02
CH-444 Added confidence measure to configuration and pass to report calls.

eshtadc:


plugins/agent_types/AcquiaLiftAgent.inc
```diff
@@ -803,6 +843,10 @@ class AcquiaLiftReport implements PersonalizeAgentReportInterface {
     $machine_name = $this->agent->getMachineName();
 
     if (!isset($this->report_data[$key])) {
+      $confidence_measure = $this->agent->getConfidenceMeasure();
```
katbailey: It feels kind of weird to pass this property through the agent as the only way of getting it into the report class. Does the agent itself actually need this property for anything? One solution, without compromising the unit testability of this class (;-)) would be to use a factory to create the report, and the factory method would call variable_get to get the confidence measure setting and then inject it into the report class's constructor. This is just a suggestion, and doesn't make sense if the agent itself does need to know about this property (or might in the future.) What do you think?

daveatacquia: I can see this making sense as part of the agent ... It would be
conceivable, for example, to have an agent which changes its behavior once
a particular significance level is reached. That's probably a bad idea, but
it's possible. Most of the time though, this would have to do with
reporting only.

On Saturday, May 3, 2014, katbailey notifications@github.com wrote:

> In plugins/agent_types/AcquiaLiftAgent.inc:
> 
> > @@ -803,6 +843,10 @@ class AcquiaLiftReport implements PersonalizeAgentReportInterface {
> >      $machine_name = $this->agent->getMachineName();
> > 
> > ```
> >  if (!isset($this->report_data[$key])) {
> > ```
> > -      $confidence_measure = $this->agent->getConfidenceMeasure();
> 
> It feels kind of weird to pass this property through the agent as the only
> way of getting it into the report class. Does the agent itself actually
> need this property for anything? One solution, without compromising the
> unit testability of this class (;-)) would be to use a factory to create
> the report, and the factory method would call variable_get to get the
> confidence measure setting and then inject it into the report class's
> constructor. This is just a suggestion, and doesn't make sense if the agent
> itself does need to know about this property (or might in the future.) What
> do you think?
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/acquia/acquia_lift/pull/19/files#r12256856
> .

## 

_Dave Ingram,_ Solutions Architect
_Mobile: *+1-352-214-1144     *Office:_ +1-781-313-8305
*Skype: *dave.acquia     *Twitter: *@ https://twitter.com/daveaingram
daveaingram https://twitter.com/daveaingram
http://www.acquia.com

http://www.linkedin.com/in/daveaingram

katbailey: That wouldn't require knowing the confidence measure used for reporting, but simply a boolean indicating whether confidence had been reached or not. And that would have to come from the report anyway.

eshtadc:
I'm alright with the factory usage here given the need for that specific variable... Typically I over-think so I was trying to keep things simple, but given that reports are still a moving target, we don't really know what other kind of oddity data we may end up needing.

eshtadc:
Refactored now to utilize a factory for report creation - also delayed the instantiation of the reporting within the Acquia Lift agent since it's only needed in a few cases.

PR #20, opened 2014-05-08
CH-458: Use new preview flag in execute call.

eshtadc:


eshtadc:
See https://github.com/acquia/personalize_module/pull/25

PR #21, opened 2014-05-19
Ch 488: Update Lift class to indicate autopersonalization with fixed targeting limited to auto-targeting contexts.

eshtadc:


eshtadc:
merged

PR #22, opened 2014-05-21
IA link path changes.

eshtadc:
Remove default Acquia Lift page in favor of updated personalize campaigns listing.


eshtadc:
Merged manually.

PR #23, opened 2014-06-12
CH-511: Update status retrieval when creating a new Lift agent.

eshtadc:
Can no longer read the status from the agent data, so need to retrieve the status when creating a new agent instance.


eshtadc:
committed

PR #24, opened 2014-06-30
Ch 561: Create a unified navigation to be integrated across menu modules.

eshtadc:
The following directories are just straight additions from navbar:
- css/navbar
- css/icons
- css/images
- js/navbar
  Each one has a README.md file to explain.


js/acquia_lift.personalize.js
```diff
@@ -159,19 +171,48 @@ Drupal.behaviors.acquiaLiftPersonalize = {
                   // Build a view for campaign goals.
                   if (type === 'campaigns') {
                     var $goalsMenu = $('[data-acquia-lift-personalize-type="goals"]');
-                    looper(model.get('goals') || {}, function (obj, key) {
+                    var modelGoals = model.get('goals');
+                    if (modelGoals) {
+                      looper(model.get('goals') || {}, function (obj, key) {
```
katbailey: Why not just `looper(modelGoals, function...)`?

js/acquia_lift.unified_navbar.admin_menu.js
```diff
@@ -0,0 +1,112 @@
...
+    // Helper method to get admin menu container,
+    // override it to provide support of custom admin menu.
+    // By default supports https://drupal.org/project/admin_menu, https://drupal.org/project/admin
+    // and default toolbar core module.
+    // https://drupal.org/project/navbar is supported @see acquia_lift.navbar.js
+    getAdminMenu: function () {
+      return $('#admin-toolbar, #admin-menu, #toolbar');
+    }
+  }
+  var self = Drupal.behaviors.acquiaLiftUnifiedNavbarIntegration;
```
katbailey, line 110: This seem out of place..? Or is it to avoid having to do `var self = this;` in the methods above?

js/acquia_lift.personalize.js
eshtadc: Hmmm - smells of bad copy/paste - yes should looper (modelGoals)... fixing it now.

js/acquia_lift.unified_navbar.admin_menu.js
eshtadc, line 110: Both to avoid multiple setting of self but also because the updateUnifiedToolbarPosition can be called in different scopes and wanted to reference the getAdminMenu from it.  Incidentally, the getAdminMenu is included in both admin_menu and toolbar integration JS files.  I'm going to update them to be specific to their integration rather than having the more generic handler.

eshtadc:
Merged.

PR #25, opened 2014-07-14
Ch 565: Create a campaign via simplified modal view

eshtadc:


plugins/agent_types/AcquiaLiftAgent.inc
```diff
@@ -200,9 +200,33 @@ class AcquiaLiftAgent extends PersonalizeAgentBase implements PersonalizeAgentGo
   }
 
   /**
+   * Gets a simplified version of the optionsForm for use when
+   * creating a simple A/B test.
+   */
+  public static function simpleABOptionsForm($agent_data) {
```
katbailey, line 210: I think eventually this needs to be broken out into a different agent type, i.e. a separate class for the "Simple AB test". Because we'll need to do other things like limiting it to just one variation set. But that's beyond the scope of this story, so maybe just add a todo about eventually moving this into its own class?

eshtadc:
merged

PR #28, opened 2014-07-23
Separate agent type for Simple A/B

katbailey:


acquia_lift.module
```diff
@@ -89,6 +89,32 @@ function acquia_lift_menu() {
 }
 
 /**
+ * Implements hook_menu_alter().
+ */
+function acquia_lift_menu_alter(&$items) {
```
eshtadc: I had no idea you could do this.  That's perfect and exactly what I should have done for changing the way the 'add' link works from the Acquia Lift menu rather than the personalize module.

acquia_lift.module
```diff
@@ -305,26 +331,53 @@ function acquia_lift_cron_queue_info() {
...
+
+/**
+ * Returns whether or not the passed in agent is owned by this module.
+ *
+ * @param $agent
+ *   An object representing the agent.
+ * @return bool
+ *   TRUE if the object is an Acquia Lift owned agent, FALSE otherwise.
+ */
+function acquia_lift_owns_agent($agent) {
```
eshtadc, line 353: Looks like some of the places where this is called were previously checking to "!$agent".  Perhaps you should just return FALSE if $agent is empty as well.  See line 370 and 395 below.

katbailey, line 353: Yep, added this.

katbailey:
Merged

PR #30, opened 2014-07-28
Ch 666

katbailey:


acquia_lift_profiles/acquia_lift_profiles.install
```diff
@@ -0,0 +1,106 @@
...
+function acquia_lift_profiles_install() {
+  $menu_name = 'acquia-lift-controls';
+  // Create a link for accessing the Profiles web admin.
+  $item = array(
+    'link_title' => 'Acquia Lift Web Admin',
+    'link_path' => 'http://lift.acquia.com',
+    'menu_name' => $menu_name,
+    'options' => array(
+      'attributes' => array(
+        'class' => array('visitor-actions-ui-ignore'),
```
eshtadc: Adding the following class will make it look like a link in Unibar: acquia-lift-menu-link

eshtadc: Hmmm - right now Unibar is set up that everything to the right is specific to the selected campaign.  This is a more general link which would put it to the left, but I'm not sure that we want it to be the first link either.  James may have something to say here....

katbailey:
Merged

PR #32, opened 2014-08-06
Ch 729

katbailey:


acquia_lift.module
```diff
@@ -539,6 +539,209 @@ function acquia_lift_sync_option_sets($agent, $old_option_sets, $new_option_sets
 }
 
 /**
+ * Batch syncs all campaigns to Lift.
```
eshtadc: This seems like code that would be rarely ever needed in regular operations of Acquia Lift so would it make sense to go to it's own file?  I guess I still have a hard time with the ginormous files in Drupal modules :p

acquia_lift.module
```diff
@@ -580,13 +783,14 @@ function acquia_lift_personalize_goal_save($goal_info) {
  *   - method The method to call on the AcquiaLiftAPI instance
  *   - args The args to pass to the method.
  */
```
eshtadc: Add $errors to docblock (even though this one is pretty obvious) - some mornings my coffee isn't strong enough....

acquia_lift.module
```diff
@@ -539,6 +539,209 @@ function acquia_lift_sync_option_sets($agent, $old_option_sets, $new_option_sets
...
+    $context['sandbox']['progress']++;
+  }
+  else {
+    // We allow one retry per operation.
+    if (!$context['sandbox']['retries']) {
+      $context['finished'] = 0.5;
+      $context['sandbox']['retries'] = 1;
+    }
+    else {
+      // Is there a better way to signal that a particular operation was
```
eshtadc: Looks like simpletest indicates a message and passes a specific div to be styled as an error in order to achieve this when running tests.
https://github.com/boombatower/drupal/blob/master/modules/simpletest/simpletest.module#L194

Not saying you need to go this far at all, but it's all I see that can indicate an error during batch processing itself.

plugins/agent_types/AcquiaLiftAgent.inc
```diff
@@ -364,6 +364,21 @@ class AcquiaLiftAgent extends PersonalizeAgentBase implements PersonalizeAgentGo
    * Implements PersonalizeAgentInterface::postSave().
    */
   public function postSave($old_data) {
+    $items = $this->getAgentSyncOperations(isset($old_data->data['visitor_context']['acquia_lift_context']));
+
+    $this->queueItems($items);
+  }
+
+  /**
+   * Returns the operations needed to sync an age t to Acquia Lift.
```
eshtadc: I swear I feel like I'm only good at code reviewing comments ;-)
"t"

plugins/agent_types/AcquiaLiftAgent.inc
```diff
@@ -751,13 +796,30 @@ class AcquiaLiftAgent extends PersonalizeAgentBase implements PersonalizeAgentGo
...
 
   /**
    * Implements AcquiaLiftAgentInterface::syncFixedTargeting().
    */
   public function syncFixedTargeting($option_sets) {
+    $items = $this->getFixedTargetingSyncOperations($option_sets);
+    $this->queueItems($items);
+  }
+
+  public function getFixedTargetingSyncOperations($option_sets) {
```
eshtadc: docblock (like I said)

plugins/agent_types/AcquiaLiftAgent.inc
```diff
@@ -364,6 +364,21 @@ class AcquiaLiftAgent extends PersonalizeAgentBase implements PersonalizeAgentGo
    * Implements PersonalizeAgentInterface::postSave().
```
eshtadc, line 364: I love the changes to this file!

tests/acquia_lift.test
```diff
@@ -907,6 +906,286 @@ class AcquiaLiftWebTest extends DrupalWebTestCase {
...
+          'first-goal'
+        )
+      ),
+    );
+    variable_set('acquia_lift_web_test_data', $test_data);
+
+    // Now try syncing this agent to Lift.
+    $agents = personalize_agent_load_multiple(array('first-existing-agent'));
+    // Ensure the test API client is used during the call for operations.
+    AcquiaLiftAPI::setTestInstance();
```
eshtadc, line 1001: I thought this was automatic... getting more coffee.

acquia_lift.module
katbailey: Yeah, good point - I could move this stuff into a .batch.inc file

acquia_lift.module
katbailey: Yeah, I guess that would be nice to do, but I was more concerned with being able to flag particular operations during processing, which could then be picked out at the end for display. Not necessarily displaying them _during_ processing. I'm going to be lazy and leave it as is for now :P

plugins/agent_types/AcquiaLiftAgent.inc
katbailey: Haha :) That's exactly how I feel when reviewing some of your monster PRs ;)

tests/acquia_lift.test
katbailey, line 1001: It's automatic when there's an actual page request being made because it uses drupal_valid_test_ua() which just checks the 'HTTP_USER_AGENT' header. But a direct function call from within a test will not result in the test instance being used. I found this out by running it :)

PR #40, opened 2014-08-27
CH-744: Add in the build process for js and css

Zarabadoo:
This will roll in the js build that i had been working with previously to break up the report graphs into more manageable pieces. This also rolls in the grunt build for the css. As a positive side effect, many unneeded vendor prefixes are removed from other css files thanks to Autoprefixer.


eshtadc:
So what does this mean now for maintenance to the module moving forward.  Would there have to be an additional step to ensure that tasks are run?

Zarabadoo:
It is all run with grunt while you are developing and commit the resulting files. Existing tests should not be affected. However, any changes to the functionality should happen the src directory and then processed with grunt.

For development, you would need Node and Ruby installed. Running `npm install` and `bundle install` in the module directory will install all the dependencies. Run `grunt` will do a build of the resources and `grunt watch` start a task that would automatically build when files in the src directory are changed.

Currently the build is only needed when touching the reports.ab related functionality and styles. If the process is deemed good, we can expand on it with other files in the module if there are items that could stand to be broken up. I mainly did it with these specific items because we have 4 Rickshaw extensions and then the jquery wrapper. Individually they are rather large files, so it is easier to parse them in smaller chunks. I have also enabled sourcemaps so we can we can see where errors are occurring in the web inspector console.

css/acquia_lift.report.ab.css
```diff
@@ -3,27 +3,32 @@
  * ----------------------------------------------------------------------------
  */
 .lift-graph {
-  font-family: sans-serif;
-}
+  font-family: sans-serif; }
```
eshtadc: I love that autoprefixer is fixing these.  Do you know a way to tell it to add the ending curly brace on a new line so that it follows the rest of the coding style in the file?  I looked through the options on github but didn't see anything.  Obviously a minimal concern compared to the benefit provided....

Zarabadoo: I did not properly specify the output on the sass processor. I just pushed up a fix.

PR #44, opened 2014-09-02
CH-832: Allow for multiple sets of data.

Zarabadoo:
This applies the needed markup changes along with javascript function needed to allow for multiple sets of data to be used.


acquia_lift.admin.inc
```diff
@@ -618,7 +618,19 @@ function acquia_lift_report_ab($form, &$form_state, $agent, $agent_data) {
   $date_end_report = $date_end_targeting = date('Y-m-d');
   // Acquia Lift A/B reports have a decision name that is the same as the
   // campaign machine name.
-  $reports = $agent->buildCampaignReports($agent_data->machine_name, $date_start_report, $date_end_report);
+  $reports = array(
```
eshtadc: I have actually made these changes while adding in the new tables.  You can see them at https://github.com/acquia/acquia_lift/commit/7e8e9ba17502d7bf3b18439fa75228b3a00ed2dd

I went ahead and updated to match the style you used here so that we should be in sync with the new reports.

js/acquia_lift.reports.js
```diff
@@ -7,11 +7,121 @@
...
+
+        // Process the data in all tables.
+        $categories.each(function (graphKey) {
+          var $this = $(this),
+              $graph = $this.find('table[data-lift-statistics]'),
+              dataLabels = [],
+              name = $this.children('.lift-statistic-category-name').addClass('element-invisible').text(),
+              variationColumn = $graph.attr('data-liftgraph-columnname') - 1,
+              xColumn = $graph.attr('data-liftgraph-columnx') - 1,
+              attr = typeof $graph.attr('data-liftgraph-excluded') !== 'undefined' ? $graph.attr('data-liftgraph-excluded').split(',') : [],
```
eshtadc, line 41: Ah - so this is the exclusion data attributes.  Looks like it expects a list of column numbers starting at column1.  Can you add a little description/comment in here about it so that it isn't missed.  It's going to be an important part for someone just putting together a table that may not be as front-endy.

PR #45, opened 2014-09-03
Ch 830: Ability to split reports by goal and by metric

eshtadc:


includes/acquia_lift.classes.inc
```diff
@@ -1469,10 +1470,11 @@ class AcquiaLiftReportDataFromFile implements AcquiaLiftReportDataSourceInterfac
    */
   public function getConfidenceReport($agent_name, $date_start = NULL, $date_end = NULL, $point = NULL, $options = array()) {
     $report_name = 'confidence';
-    // Note that the value is the string false for use in JSON rather than
-    // the boolean value.
+    if (!empty($options['goal'])) {
+      $report_name .= '_goal' . rand(1, 2);
```
katbailey, line 1468: The use of rand here could use an explanation.

plugins/agent_types/AcquiaLiftAgent.inc
```diff
@@ -1031,6 +1031,38 @@ abstract class AcquiaLiftReportBase implements PersonalizeAgentReportInterface,
...
+   *
+   * @returns array
+   *   An array of confidence report options.
+   */
+  function getConfidenceDetailReportOptions() {
+    $options['aggregated-over-dates'] = FALSE;
+    return $options;
+  }
+
+  /**
```
katbailey, line 1077: I think the above two methods should be declared protected.

plugins/agent_types/AcquiaLiftAgent.inc
```diff
@@ -1284,14 +1295,64 @@ class AcquiaLiftABReport extends AcquiaLiftReportBase {
...
+   *   True if the data already exists
+   *   False if the data loaded in error.
+   */
+  protected function loadConfidenceReportHelper($report_data, $detail, $options = array()) {
+    if ($detail) {
+      $options = array_merge($options, $this->getConfidenceDetailReportOptions());
+    }
+    $report_name = $this->getConfidenceReportRawName($options);
+    if (isset($report_data['raw'][$report_name])) {
+      return TRUE;
```
katbailey, line 1352: I know we already have lots of methods where we return _either_ an array _or_ boolean FALSE if nothing can be returned but that's generally not good practice and we really shouldn't do it. And here we return an array _or_ TRUE _or_ FALSE. I know this is just a helper method, not part of the public interface, and I'm being overly picky - but it seems we could just refactor things slightly so that the calling code gets the report name and checks for its existence in the report data before calling this method. Maybe with the merging of options it becomes trickier, so feel free to respond with a big "MEH!" ;)

PR #46, opened 2014-09-05
CH-855. Allow libraries module to consider 2000 lines of underscore-min....

alanevans:
...js to attempt to find the version


katbailey:
I created 2 separate d.o issues for these fixes and committed them:
Underscore issue: https://www.drupal.org/node/2334107
Personalized fields issue: https://www.drupal.org/node/2334109

PR #48, opened 2014-09-08
CH-901 by Zarabadoo: Add qtip tooltips to statistic table headers.

Zarabadoo:


js/acquia_lift.help.js
```diff
@@ -0,0 +1,46 @@
...
+  Drupal.behaviors.acquiaLiftHelp = {
+    attach: function (context, settings) {
+      $('[data-help-tooltip]').each(function() {
+        $(this).once('acquia-lift-help', function() {
+          var text = $(this).attr('data-help-tooltip'),
+              trigger = $('<span class="acquia-lift-help">'),
+              positionTooltip = 'bottomMiddle';
+
+          // The tooltip will run out of room on the right.
+          if ($(this).is(':last-child')) {
```
eshtadc, line 20: Is this just because we are using it right now with table column headers?  If so, perhaps it is better served in something specific to the A/B report rather than the generic?  Of course, we could just cross that bridge when we come to it and just add a quick note.  And then there is the case that I just have complete misunderstood ;-)

plugins/agent_types/AcquiaLiftAgent.inc
```diff
@@ -1644,11 +1644,26 @@ class AcquiaLiftABReport extends AcquiaLiftReportBase {
...
+        'data' => t('Chance to beat control'),
+        'data-help-tooltip' => t('Likelihood visitors will complete goals for a variation compared to the control.'),
+      ),
+      array(
+        'data' => t('Lift'),
+        'data-help-tooltip' => t('Likelihood visitors will complete goals for a variation compared to a random variation.'),
+      ),
+      array(
+        'data' => t('Winner'),
+        'data-help-tooltip' => t('Most effective variation for visitors based on a 95% confidence level.'),
```
eshtadc, line 1665: The 95 needs to be filled in with a call to getConfidenceMeasure() as it may not always be 95%.

js/acquia_lift.help.js
eshtadc, line 20: Darn it - pretend this comment is on the src js rather than the compiled js.

PR #56, opened 2014-09-18
CH-921: Make the list of campaigns scrollable when the orientation is horizontal

Zarabadoo:


js/acquia_lift.personalize.js
```diff
@@ -137,6 +137,13 @@
                     });
                     $('[data-acquia-lift-personalize-type="campaigns"]').prepend(element);
                   }
+                  else {
+                    // If there are campaigns, create a container to hold them.
+                    $menu = $('[data-acquia-lift-personalize-type="campaigns"]');
+                    scrollable = document.createElement('ul');
```
eshtadc: I think this is missing a "var".

js/acquia_lift.personalize.js
```diff
@@ -137,6 +137,13 @@
                     });
                     $('[data-acquia-lift-personalize-type="campaigns"]').prepend(element);
                   }
+                  else {
+                    // If there are campaigns, create a container to hold them.
```
eshtadc: Can you elaborate somehow in this comment that the reason a separate container is to keep that "add link" at the bottom visible?

js/acquia_lift.personalize.js
```diff
@@ -207,6 +214,7 @@
         // top-level links in the Acquia Lift menu.
         _.each(['campaigns', 'option_sets'], function (category) {
           var $typeMenus = $('[data-acquia-lift-personalize-type="' + category + '"]');
+          var $scrollable = $typeMenus.siblings('.acquia-lift-scrollable');
```
eshtadc, line 219: Maybe it would be less error-prone down the way if there was a common variable here - something like:
var $holder = $scrollable.length > 0 ? $scrollable : $menu;

And then on 265 - you don't need the wrapping if and it can just be $parent.prepend(element) -- and hopefully we're less likely to screw up that logic in the future.

(Not a necessary change - just a thought).

Zarabadoo, line 219: That is a solid idea and implemented.

js/acquia_lift.personalize.js
Zarabadoo: I am not sure how I missed that... or why it was working in the first place.

js/acquia_lift.personalize.js
Zarabadoo: New expanded comment added.

Zarabadoo:
This has been merged.

PR #57, opened 2014-09-23
Ch 904

katbailey:


acquia_lift.module
```diff
@@ -614,8 +662,15 @@ function acquia_lift_personalize_goal_save($goal_info) {
  *   - args The args to pass to the method.
  */
 function acquia_lift_sync_item($item) {
-  $acquia_lift_api = AcquiaLiftAPI::getInstance(variable_get('acquia_lift_account_info', array()));
-  call_user_func_array(array($acquia_lift_api, $item['method']), $item['args']);
+  // The item is either a method to be called on the LiftAPI class or a regular
```
eshtadc, line 677: Is it possible to combine the method and callbacks into a callable somehow so that there is uniformity in how it is handled?
http://php.net/manual/en/language.types.callable.php

acquia_lift.module
```diff
@@ -2058,6 +2113,45 @@ function acquia_lift_form_field_ui_field_edit_form_alter(&$form, &$form_state, $
       'valid' => array(':input[name="field[cardinality]"]' => array('value' => -1)),
     );
     $form['field']['settings']['personalize']['enabled']['#description'] = t('This setting is only valid for fields with unlimited cardinality.');
+    // Add a state to the form elements to make sure they only appear when the
+    // personalizable checkbox is checked.
+    $state = array(':input[name="field[settings][personalize][enabled]"]' => array('checked' => TRUE));
+
+    // Checkbox for auto creation of a goal should only be visible if the selected
+    // campaign type supports goals.
```
eshtadc: Yes - but it's limited here to Acquia Lift agents rather than by goals interface (which I guess makes sense because they all support goals).  Could the comment either be more clear how this is working or can we do some sort of goal checking to make it follow the logic?

tests/acquia_lift.test
```diff
@@ -1568,6 +1568,268 @@ class AcquiaLiftWebTest extends DrupalWebTestCase {
...
+      'entity_type' => 'node',
+      'label' => 'Personalizable Headline',
+      'bundle' => 'article',
+      'required' => FALSE
+    );
+    field_create_instance($instance);
+
+    // Create a node which we will attach a fields-based option set to. We can't
+    // do this via the form because of the way we alter the form, which makes the
+    // "Add an option" button not findable by simpletest.
```
eshtadc, line 1613: Wow.

acquia_lift.module
katbailey, line 677: Yep, that makes much more sense - thanks, will change.

acquia_lift.module
katbailey: Yeah, copy/paste error because I had originally put this in personalize_fields module before realizing it would be way too complicated to support the logic there, will update the comment.

tests/acquia_lift.test
katbailey, line 1613: Yeah, this was a total PITA :(

acquia_lift.module
katbailey, line 677: Hmm, actually on second thought this isn't so straight-forward. These aren't static methods on a class so we can't just have something like `'callable' => array('AcquiaLiftAPI', 'saveAgent')` - you'd have to pass the actual instantiated LiftAPI object, which would then get serialized and... ugh. There's also a ton of existing code this would affect so it would be quite an upheaval. The conditional is definitely ugly but for now I think the best I can do is add a todo about figuring out a way to handle it better.

PR #59, opened 2014-09-26
Ch 955

katbailey:


acquia_lift.admin.inc
```diff
@@ -115,7 +115,12 @@ function acquia_lift_admin_form($form, &$form_state) {
     '#description' => t("Paste in your Acquia Lift API URL"),
     '#required' => TRUE,
   );
-
+  $form['acquia_lift_batch_decisions'] = array(
+    '#type' => 'checkbox',
+    '#default_value' => variable_get('acquia_lift_batch_decisions', FALSE),
```
eshtadc, line 120: So is this another variable to delete upon uninstall then?

katbailey, line 120: Yes indeed! Just added that, as well as the acquia_lift_confidence_measure var which also wasn't being deleted.

eshtadc, line 120: Yeah - when I thought of it I figured there would be more.  It's one of those things that's too easy to forget about.

PR #62, opened 2014-10-02
Ch 986: Goals queue

eshtadc:


js/acquia_lift.api.js
```diff
@@ -0,0 +1,99 @@
+/**
+ * Contains the definition of the Acquia Lift API.
+ *
+ * The abstraction of the API wrapper into it's own singleton allows different
```
katbailey: For the sake of having something to comment on, I'm going to point out a stray apostrophe - should be "into its own..." You're welcome :)

js/acquia_lift.goals_queue.js
```diff
@@ -0,0 +1,467 @@
...
+      queueItemData,
+      queueItemProcessing = false;
+
+    /**
+     * Returns the unique ID assigned to this queue item.
+     */
+    this.getId = function () {
+      return queueItemUid;
+    };
+    this.setId = function (value) {
```
katbailey, line 26: Missing doc block

js/acquia_lift.goals_queue.js
```diff
@@ -0,0 +1,467 @@
...
+      queueItemUid = value;
+    }
+
+    /**
+     * Returns the data that is held by this queue item.
+     */
+    this.getData = function () {
+      return queueItemData;
+    };
+    this.setData = function (value) {
```
katbailey, line 40: Again - or is that intentional? Just lump the getter and setter together with one docblock?

js/acquia_lift.goals_queue.js
```diff
@@ -0,0 +1,467 @@
...
+
+    // Constructor handling.
+    if (params.hasOwnProperty('id')
+      && params.hasOwnProperty('data')
+      && params.hasOwnProperty('pflag')) {
+      this.setId(params.id);
+      this.setData(params.data);
+      this.setProcessing(params.pflag);
+    } else {
+      var uid = 'acquia-lift-ts-' + new Date().getTime();
```
katbailey: Seems this is not unique enough - it can result in multiple queue items with the same uid if they are processed quickly enough. Should suffice to just append some Math.random() to it.

js/acquia_lift.goals_queue.js
eshtadc, line 26: Was struggling with a scope issue when I made the setters... just missed adding their own docblocks.

js/acquia_lift.api.js
eshtadc: lol

PR #71, opened 2014-10-23
CH-1120: Use Drupal's element-hidden class to hide the variation edit icon

Zarabadoo:
Currently, the icon is hidden with jquery's toggle(). This is problematic in that it adds inline style `display: inline;` and overrides the css.

### DIFF
--- a/js/acquia_lift.personalize.js
+++ b/js/acquia_lift.personalize.js
@@ -1798,7 +1798,7 @@
         this.$el
           .toggleClass('acquia-lift-page-variation-toggle-disabled', currentCampaign.get('activeVariation') == 0) // There is no toggle available for the control variation.
           .toggleClass('acquia-lift-page-variation-toggle-active', this.model.get('isActive'))
-          .toggleClass('element-hidden', currentCampaign instanceof Drupal.acquiaLiftUI.MenuCampaignABModel);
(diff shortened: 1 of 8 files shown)


js/acquia_lift.personalize.js
```diff
@@ -1798,7 +1798,7 @@
         this.$el
           .toggleClass('acquia-lift-page-variation-toggle-disabled', currentCampaign.get('activeVariation') == 0) // There is no toggle available for the control variation.
           .toggleClass('acquia-lift-page-variation-toggle-active', this.model.get('isActive'))
-          .toggle(currentCampaign instanceof Drupal.acquiaLiftUI.MenuCampaignABModel);
+          .toggleClass('element-hidden', currentCampaign instanceof Drupal.acquiaLiftUI.MenuCampaignABModel);
```
eshtadc: There are a couple of problems unfortunately.  The first is easy - this is backwards ;-)  The element-hidden class would be applied if the currentCampaign is not an instance of the MenuCampaignABModel class.

The harder issue is that when you add the element-hidden class, it doesn't actually hide the element because the display: none of element-hidden is overridden by the display: inline-block of the acquia-lift-page-variation-toggle class.

Zarabadoo: I will reverse the logic. The display is a non-issue as `.element-hidden` is using `!important` and overrides `.acquia-lift-page-variation-toggle` since it does not use it.

PR #74, opened 2014-10-27
Ch 1126

katbailey:


acquia_lift.module
```diff
@@ -537,14 +537,23 @@ function acquia_lift_personalize_option_set_save($option_set) {
     $option_set->is_new = TRUE;
   }
   // Unless this is a new fields-based option set, we're done.
```
eshtadc: Not anymore....

eshtadc:
This test definitely looks like it was far more of a pain to figure out than the problem was to solve.

PR #79, opened 2014-11-03
Ch 1143: Allow renaming of options from unibar

eshtadc:


acquia_lift.admin.inc
```diff
@@ -1328,13 +1328,127 @@ function acquia_lift_page_variation_rename_form($form, &$form_state, $agent_name
...
 
 /**
+ * Page callback to generate a option set's option label rename modal window.
+ *
+ * @param string $osid
+ *   The option set id string that contains the option to rename.
+ * @param string $option_id
+ *   The option id of the option to rename.
+ */
+function acquia_lift_option_rename_callback($osid_str, $option_id) {
```
katbailey: For clarity could we call this function acquia_lift_option_rename_modal_callback()? Makes it clearer that it's responsible for the modal, not for the actual renaming.

eshtadc: Sure.  There are a few function renames then for consistency.  There were a few others I contemplated making anyway... so I'll just go ahead and do them all in a follow-up commit.

PR #80, opened 2014-11-04
CH-1142/CH-1178: Add the agent type querystring param to the "Custom Lif...

katbailey:
...t campaign" link

This includes some variable renaming while I was in the theme function and also I think a bug fix which I will point out in the code.

### DIFF
--- a/theme/acquia_lift.theme.inc
+++ b/theme/acquia_lift.theme.inc
@@ -276,6 +276,7 @@ function theme_acquia_lift_type_list($variables) {
   ctools_include('modal');
   ctools_include('ajax');
   ctools_modal_add_js();
+  drupal_add_library('system', 'drupal.ajax');
 
   $items = $variables['items'];
   $html = '<div';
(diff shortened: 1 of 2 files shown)


theme/acquia_lift.theme.inc
```diff
@@ -283,24 +284,38 @@ function theme_acquia_lift_type_list($variables) {
...
-  foreach ($items as $typestr => $current_type) {
+  foreach ($items as $typestr => $type) {
     $html .= '<div class="acquia-lift-type clearfix">';
-    $html .= '<div class="acquia-lift-type-logo">' . $current_type['logo'] . '</div>';
+    $html .= '<div class="acquia-lift-type-logo">' . $type['logo'] . '</div>';
     $html .= '<div class="acquia-lift-type-title">';
-    if ($current_type['modal']) {
-      $url_parts = explode('/', $current_type['url']);
-      end($url_parts);
-      if (key($url_parts) != 'nojs') {
```
katbailey, line 293: Pretty sure this was a bug that we just never saw. key() would have returned the array key, not the value.

eshtadc, line 293: hmmmm yeah it must have been.  Perhaps all of the times this is called the url is already properly set up.

PR #81, opened 2014-11-05
Ch 1144: Allow renaming goals from unibar

eshtadc:


acquia_lift.admin.inc
```diff
@@ -1848,6 +1848,91 @@ function acquia_lift_goal_type_create_form_submit($form, &$form_state) {
 }
 
 /**
+ * Page callback to generate a goal rename modal window.
+ *
+ * @param array $action
+ *   The visitor action for the goal.
```
katbailey, line 1854: This doesn't match the actual param.

acquia_lift.ui.inc
```diff
@@ -65,6 +65,17 @@ function acquia_lift_build_page(&$page) {
     $page['page_top']['#attached']['library'][] = array('acquia_lift', 'acquia_lift.modal');
     ctools_add_css('acquia_lift.ctools.modal', 'acquia_lift');
 
+    // Load data about which visitor actions are custom-defined versus those
+    // defined in code.
+    $actions = visitor_actions_get_actions();
+    $actions_defined = array();
```
katbailey, line 71: I think you just need visitor_actions_custom_load_multiple() here to get all actions that are in the db as opposed to defined in code.

eshtadc, line 71: Yeah - I think I was thinking ahead (to what I don't know) and over-complicating.

PR #84, opened 2014-11-11
Ch 1147: Allow deletion of goals

eshtadc:
Includes the deletion of goals but also a maintenance task to split the acquia_lift.personalize.js file into several smaller js files for manageability.  You can see the actual changes made to the individual files in the commit history rather than the full file change history.


acquia_lift.admin.inc
```diff
@@ -1938,6 +1938,105 @@ function acquia_lift_goal_rename_form_submit($form, &$form_state) {
...
+ *   The machine name for the visitor action for the goal.
+ */
+function acquia_lift_goal_delete_form($form, &$form_state, $agent_name, $action_name) {
+  ctools_include('modal');
+  ctools_include('ajax');
+  ctools_add_js('ajax-responder');
+
+  $form = array();
+  // Will only load if this is a custom action (not defined in code).
+  $action = visitor_actions_custom_load($action_name);
```
katbailey, line 1992: Should probably check that an action was returned and otherwise output an error message rather than the form.

eshtadc, line 1992: Well we can delete a goal that is from a visitor action defined in code so it is still valid.  We just load this here to get the label for the action.

katbailey, line 1992: Ah right, that makes sense.

PR #85, opened 2014-11-17
Issue #2376775 by eshta: Fix Acquia Lift Web Admin link href to be dynam...

eshtadc:
...ically read from API url configuration setting.

### DIFF
--- a/acquia_lift_profiles/acquia_lift_profiles.module
+++ b/acquia_lift_profiles/acquia_lift_profiles.module
@@ -25,6 +25,19 @@ function acquia_lift_profiles_menu() {
 }
 
 /**
+ * Implements hook_translated_menu_link_alter().
+ */
+function acquia_lift_profiles_translated_menu_link_alter(&$link) {
+  // Update the admin link if the API url is specified in the config.
+  if ($link['menu_name'] === 'acquia-lift-controls' && $link['link_path'] === 'http://lift.acquia.com') {
+    $updated_link = 'http://' . variable_get('acquia_lift_profiles_api_url', 'lift.acquia.com');
+    if (valid_url($updated_link, TRUE)) {
+      $link['href'] = $updated_link;
+    }
+  }
+}
+
+/**
  * Implements hook_permission().
  */
 function acquia_lift_profiles_permission() {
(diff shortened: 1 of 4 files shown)


acquia_lift_profiles/tests/acquia_lift_profiles.test
```diff
@@ -264,7 +270,7 @@ class ALProfilesWebTest extends DrupalWebTestCase {
     $edit = array(
       'acquia_lift_profiles_account_name' => 'MYALPROFILESACCOUNT',
       'acquia_lift_profiles_js_path' => 'acquia_lift_profiles.example.com/path/to.js',
-      'acquia_lift_profiles_api_url' => 'api.acquia_lift_profiles.example.com',
```
eshtadc, line 267: FYI:  I changed this because api.acquia_lift_profiles.example.com doesn't return true from drupal's valid_url() check.

acquia_lift_profiles/acquia_lift_profiles.module
```diff
@@ -25,6 +25,19 @@ function acquia_lift_profiles_menu() {
 }
 
 /**
+ * Implements hook_translated_menu_link_alter().
+ */
+function acquia_lift_profiles_translated_menu_link_alter(&$link) {
+  // Update the admin link if the API url is specified in the config.
+  if ($link['menu_name'] === 'acquia-lift-controls' && $link['link_path'] === 'http://lift.acquia.com') {
+    $updated_link = 'http://' . variable_get('acquia_lift_profiles_api_url', 'lift.acquia.com');
```
katbailey, line 33: We need to be sure the configured API url does not already include the scheme. Let's make a change in the admin form to strip it out if it's entered. And here we should use whatever scheme is currently being used like we do in the getInstance() method in the api client class, i.e.

```
    global $is_https;
    // Use the same scheme for Acquia Lift Profiles as we are using here.
    $url_scheme = ($is_https) ? 'https://' : 'http://';
    $api_url = $url_scheme . $api_url;
```

eshtadc, line 33: I'm adding in the part to strip it out of the config form and a test for that.  I don't think I need to change the protocol here, however, as this is checking to see if it is the default value saved in the install (which has an external link path of 'http://lift.acquia.com').  And we aren't making a call or including anything... we are simply adding a menu link to the admin.

eshtadc, line 33: Although I suppose it really should be https://lift.acquia.com, but it does get redirected properly anyway when accessed via HTTP get.

PR #89, opened 2014-11-26
Ch 1408 - Pause agents in Lift when they are paused in Drupal

katbailey:
I ended up doing a lot of work on how we handle bad responses from Lift as part of this story. The main reason for doing this was to allow the queue to be aborted if a certain type of exception is thrown. If we are making a bad request that Lift cannot handle, then the queue should not proceed to the next item, which could be setting the status to running. Instead the user should get a message that something went wrong.

### DIFF
--- a/includes/acquia_lift.classes.inc
+++ b/includes/acquia_lift.classes.inc
@@ -208,6 +208,18 @@ class AcquiaLiftAPI implements PersonalizeLoggerAwareInterface, AcquiaLiftReport
     return $code >= 500 && $code < 600;
   }
 
+  /**
+   * Maps the passed in response code to an exception class to use.
+   *
+   * If the response code passed in constitutes a successful response,
+   * NULL is returned. If it does not constitute a 400 or 500 error,
+   * NULL is returned.
+   *
+   * @param $code
+   *   The response code to map.
+   * @return null|string
+   *   The exception class to use or NULL.
+   */
   protected static function mapBadResponseToExceptionClass($code) {
     if (self::isSuccessful($code)) {
       return NULL;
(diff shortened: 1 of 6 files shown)


includes/acquia_lift.classes.inc
```diff
@@ -345,9 +442,55 @@ class AcquiaLiftAPI implements PersonalizeLoggerAwareInterface, AcquiaLiftReport
...
+    $url = $this->generateEndpoint("/agent-api/$machine_name");
+    $agent = array(
+      'status' => $lift_status_code,
+    );
+    $response = $this->httpClient()->put($url, array('Content-Type' => 'application/json; charset=utf-8', 'Accept' => 'application/json'), $agent);
+    $data = json_decode($response->data, TRUE);
+    $vars = array('agent' => $machine_name);
+    $success_msg = 'The new status of campaign {agent} was pushed to Acquia Lift';
+    $fail_msg = 'The new status of campaign {agent} could not be pushed to Acquia Lift';
+    if ($response->code == 200 && $data['status'] == 'ok') {
```
eshtadc, line 481: Use the new API static isSuccessful() here?  Looks like there are several functions in this class that could use it to pull that logic out.

includes/acquia_lift.classes.inc
```diff
@@ -610,13 +744,13 @@ class AcquiaLiftAPI implements PersonalizeLoggerAwareInterface, AcquiaLiftReport
   public function getAgent($machine_name) {
     $url = $this->generateEndpoint("/agent-api/$machine_name");
     $response = $this->httpClient()->get($url, array('Accept' => 'application/json'));
-    if ($response->code == 404) {
-      throw new AcquiaLiftNotFoundException('No such agent');
+    if ($response->code == 200) {
```
eshtadc, line 771: Hmm - I would say the same thing here (use isSuccessful()) but isn't there a returned status code of 202?  Maybe that's why you left these....

includes/acquia_lift.classes.inc
```diff
@@ -122,17 +130,105 @@ class AcquiaLiftAPI implements PersonalizeLoggerAwareInterface, AcquiaLiftReport
...
+   * @param $code
+   *   The status code.
+   * @return bool
+   *   TRUE if the code represents a server error, FALSE otherwise.
+   */
+  public static function isServerError($code) {
+    return $code >= 500 && $code < 600;
+  }
+
+  protected static function mapBadResponseToExceptionClass($code) {
```
eshtadc, line 223: It's pretty self-evident... but I have to comment on comments so it needs a docblock, lol.

includes/acquia_lift.classes.inc
katbailey, line 771: Yeah I just don't want to introduce any side effects here and currently we accept only a 200 response. I'm not sure what would cause a >200 successful response in any of these calls, so for now they'll just get logged as warnings in the handleBadResponse() method.

PR #92, opened 2014-12-04
Ch 1362: Use DOM Selector for all personalize elements

eshtadc:
Note that this is compoared to a reorg branch that included lots of file renaming and moving.  I tried to pull all relevant code into this PR so that review could take place without reading every line.


acquia_lift.admin.unibar.inc
```diff
@@ -347,15 +376,45 @@ function acquia_lift_page_variation_details_form_submit($form, &$form_state) {
   if (is_numeric($form_state['values']['variation_number']) && $form_state['values']['variation_number'] >= 0) {
     $variation_number = $form_state['values']['variation_number'];
   }
+  module_load_include('inc', 'acquia_lift', 'acquia_lift.page_variations');
   $variation_number = acquia_lift_page_variation_create($variation_set_name, $option_set, array('personalize_elements_content' => $form_state['values']['personalize_elements_content']), $variation_number);
-  // Save this for the Ajax callback.
+  // Save the updated option sets for use within the Ajax callback.
+  $form_state['storage']['option_sets'] = personalize_option_set_load_multiple(array(), array('decision_name' => $variation_set_name));
```
katbailey, line 382: Shouldn't we also be constraining the option sets by agent here? Two agents could in theory use the same decision name, right?

PR #93, opened 2014-12-08
Ch 1338 - Ensure campaigns are paused when necessary, and only then

katbailey:


acquia_lift.module
```diff
@@ -709,6 +723,25 @@ function acquia_lift_sync_option_sets($agent, $old_option_sets, $new_option_sets
 }
 
 /**
+ * Syncs fixed targeting rules to Acquia Lift.
+ * @param $agent
+ * @param $new_option_sets
+ */
+function acquia_lift_sync_fixed_targetng($agent, $new_option_sets) {
```
eshtadc: Typo in function name (and where it's called...)

acquia_lift.module
```diff
@@ -724,10 +757,34 @@ function acquia_lift_personalize_goal_save($goal_info) {
   // For the current goal, take the info from what was passed in, rather
   // than from the db.
   $new_goals[$goal_info['name']] = $goal_info['value'];
-  if ($new_goals == $old_goals) {
+  acquia_lift_sync_goals($agent, $old_goals, $new_goals, TRUE);
+}
+
+function acquia_lift_personalize_goal_delete($goal_info) {
```
eshtadc, line 768: This just implements the new goal delete hook, right?

acquia_lift.module
```diff
@@ -724,10 +757,34 @@ function acquia_lift_personalize_goal_save($goal_info) {
...
+    unset($new_goals[$goal_info['name']]);
+  }
+  acquia_lift_sync_goals($agent, $old_goals, $new_goals, TRUE);
+  // If we now have no goals, the agent needs to be paused.
+  if (count($new_goals) == 0) {
+    personalize_pause_if_running($agent->machine_name);
+  }
+}
+
+function acquia_lift_sync_goals($agent, $old_goals, $new_goals, $sync_only_on_changes = FALSE) {
```
eshtadc, line 799: docblock

acquia_lift.module
```diff
@@ -862,6 +919,7 @@ function acquia_lift_process_queue($exit_on_finish = TRUE) {
       // a request made to Lift, we should abort the processing of the queue and
       // set a message to let the user know things went wrong.
       drupal_set_message(t('An error occurred while syncing your campaign information to Lift: @error', array('@error' => $e->getMessage())));
+      AcquiaLiftQueue::handleFailedItem($item);
```
eshtadc, line 941: Should this be called in the catch for other sync operations (option sets, targeting)?

eshtadc, line 941: Or is that handled by the max retries?  And if so... could it be here too?

tests/acquia_lift.test
```diff
@@ -752,7 +761,7 @@ class AcquiaLiftWebTestVariationSets extends AcquiaLiftWebTestBase {
 
     // Verify that the user is redirected to the original page with a custom
     // URL query parameter.
-    $this->assertUrl('node', array('query' => array('pm' => 'new_block|My test block')));
+    $this->assertUrl('node', array('query' => array('liftpm' => 'new_block|My test block')));
```
eshtadc, line 764: Darn it.  Thanks!

acquia_lift.module
katbailey, line 941: Yes, this is also called in the queue class's releaseItem method when the max retries is reached and the item's being deleted from the queue.

tests/acquia_lift.test
katbailey, line 764: Actually I should do this one as a hotfix and remove it from this PR... Stupid Jenkins job is stupid :(

PR #96, opened 2014-12-11
Ch 1357 - Add a new agent type for fixed targeting to Acquia Lift module

katbailey:
For now the agent type will only show up if you have the acquia_lift_target_enabled variable set to 1.

### DIFF
--- a/js/acquia_lift_target.js
+++ b/js/acquia_lift_target.js
@@ -1,0 +1,152 @@
+(function ($) {
+
+Drupal.acquia_lift_target = (function() {
+
+  var agentRules = {}, initialized = false;
+
+  function init() {
+    var i, j, optionSet, agentName, optionId;
+    var option_sets = Drupal.settings.personalize.option_sets;
+    var agent_map = Drupal.settings.personalize.agent_map;
+    for (i in option_sets) {
+      if (option_sets.hasOwnProperty(i) && agent_map.hasOwnProperty(option_sets[i].agent) && agent_map[option_sets[i].agent]['type'] == 'acquia_lift_target') {
+        optionSet = option_sets[i];
+        agentName = optionSet.agent;
+        if (agentRules.hasOwnProperty(agentName)) {
+          continue;
+        }
+        agentRules[agentName] = {};
+        for (j in optionSet.options) {
+          if (optionSet.options.hasOwnProperty(j) && optionSet.options[j].hasOwnProperty('fixed_targeting')) {
+            optionId = optionSet.options[j].option_id;
+            agentRules[agentName][optionId] = {
+              'strategy': optionSet.options[j].fixed_targeting_strategy,
+              'features': optionSet.options[j].fixed_targeting
+            };
+          }
+        }
+      }
+    }
+  }
+
+  function contextToFeatureString(key, value) {
+    return key + '::' + value;
+  }
+
+  function convertContextToFeatureStrings(visitor_context) {
+    var i, j, feature_strings = [];
+    for (i in visitor_context) {
+      if (visitor_context.hasOwnProperty(i)) {
+        for (j in visitor_context[i]) {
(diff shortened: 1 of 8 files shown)


js/acquia_lift_target.js
```diff
@@ -0,0 +1,152 @@
...
+              default:
+                // If any of the features are present, call the callback with this option
+                // as the chosen option.
+                matched = false;
+                // Set matched to true if *any* feature is present.
+                for (j in agentRules[agent_name][optionId].features) {
+                  if (agentRules[agent_name][optionId].features.hasOwnProperty(j)) {
+                    if (feature_strings.indexOf(agentRules[agent_name][optionId].features[j]) !== -1) {
+                      matched = true;
+                      break;
```
eshtadc, line 102: Does this not cause any problems with the fact that you are breaking out of a loop within a switch?  I was going to comment that you should "break" out in the "AND" case above, but figured you kept it running for simplicity.  I know you can do both with labels, but I don't see that hear.  Of course then again... if it works, it works... but it looks a bit odd.

js/acquia_lift_target.js
```diff
@@ -0,0 +1,152 @@
...
+                  decisions[decision_name] = optionId;
+                  callback(decisions);
+                  return;
+                }
+                break;
+            }
+          }
+        }
+      }
+      callback(decisions);
```
eshtadc, line 116: Is this callback necessary?

js/acquia_lift_target.js
```diff
@@ -0,0 +1,152 @@
...
+        decisionName = i;
+        choice_array = choices[decisionName];
+        break;
+      }
+    }
+    Drupal.acquia_lift_target.getDecision(agent_name, visitor_context, choice_array, decisionName, fallbacks, callback);
+  },
+  'sendGoalToAgent': function(agent_name, goal_name, value) {
+    if (window.console) {
+      console.log('Sending goal ' + goal_name + ' to agent ' + agent_name + ' with value ' + value);
```
eshtadc, line 140: This should call your fancy debugger :-)
And so should the relevant decision calls, unless the callback is handling it there....

qunit/acquia_lift_target.tests.js
```diff
@@ -0,0 +1,125 @@
...
+  Drupal.personalize.agents.acquia_lift_target.getDecisionsForPoint(agentName, evaluatedVisitorContexts, choices, decisionName, fallbacks,     callback = function(decisions) {
+    assert.ok(decisions.hasOwnProperty(decisionName));
+    assert.equal(decisions[decisionName], 'third-option');
+  });
+});
+
+/**
+ * Adds settings for the required targeting agent set-up to Drupal.settings.
+ */
+function addLiftTargetToDrupalSettings(agent_name, enabled_contexts, decision_name, osid, options) {
```
eshtadc, line 91: This only gets called once right now... but it seems like as tests expand, it could cause confusion that the agent and option info is appended to the existing information instead of overwriting it.

js/acquia_lift_target.js
katbailey, line 102: Hmm, actually I think there should be a break in the "AND" case as well, right after `matched = false;`
In both cases it just terminates the for loop so wouldn't affect the switch statement. Apparently labels are to be avoided as they are even more confusing https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/label

js/acquia_lift_target.js
katbailey, line 116: Yes, in case no rules are found, so the fallbacks will be used.

js/acquia_lift_target.js
katbailey, line 140: Ah, I had copied this from personalize_target's js and it doesn't support goals. We will be supporting goals in this agent plugin eventually but for now I'm just replacing this with a todo about that. For the debugger this is handled more centrally in personalize.js.

PR #99, opened 2014-12-18
Ch-1100 Truncate campaign and variation names

Zarabadoo:


src/css/acquia_lift.navbar.scss
```diff
@@ -0,0 +1,357 @@
+/**
```
eshtadc, line 1: Thanks for taking the opportunity to pull this into scss :-)  Now I have a decent place to start.

Zarabadoo:
A bug, needs to be merged into a different branch.

PR #101, opened 2014-12-21
Ch 1401: Adding ability to edit existing element option set options

eshtadc:


acquia_lift.admin.unibar.inc
```diff
@@ -415,14 +418,28 @@ function _acquia_lift_personalize_element_variation_details_form_submit($form, &
...
+    }
+    else {
+      // Editing an existing option set option.
+      foreach ($option_set->options as &$option) {
+        if ($option['option_id'] === $values['variation_number']) {
+          $option['personalize_elements_content'] = $values['personalize_elements_content'];
+          break;
+        }
+      }
+      $message = t('The variation has been edited.');
```
katbailey: Probably want to say "updated" here, but we can leave it for mid-sprint review. And yes, I did only add this comment for the sake of having something to say :)

PR #104, opened 2015-01-12
Ch 1703: Don't show report charts/tables when decisions hasn't been shown.

eshtadc:
Note: the two overview report function chunks are just moving the functions from the Lift report class down to the base class for use with both lift report types.

### DIFF
--- a/plugins/agent_types/AcquiaLiftAgent.inc
+++ b/plugins/agent_types/AcquiaLiftAgent.inc
@@ -1626,6 +1626,50 @@ abstract class AcquiaLiftReportBase implements PersonalizeAgentReportInterface, 
   }
 
   /**
+   * Extracts the required overview data from the report data returned by
+   * Acquia Lift.
+   *
+   * @param $items
+   *   An array of items as returned from Acquia Lift.
+   * @return array
+   *   An associative array with information for today's and the total overview.
+   */
+  protected function extractOverviewReportData($items) {
+    $agent_data = $this->agent->getData();
+    $decision_style = $agent_data['decision_style'] === 'adaptive' ? t('Auto-personalize') : t('A/B');
+    $total_variations = count($agent_data['decisions']);
+
+    // Create an array of data for each time option.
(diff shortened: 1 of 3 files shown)


plugins/agent_types/AcquiaLiftAgent.inc
```diff
@@ -2248,47 +2319,49 @@ class AcquiaLiftReport extends AcquiaLiftReportBase {
...
+            'conversion' => $choice['conversion'],
+          );
+          $option_numbers[$feature_string . '|' . $choice['raw_label']] = $choice['counter'];
+        }
       }
     }
 
     // Build the experiment report data
-    $report_data['experiment']['choices'] = $report_data['confidence']['features'][self::NO_FEATURES];
+    $report_data['experiment']['choices'] = isset($report_data['confidence']['features'][self::NO_FEATURES]) ? $report_data['confidence']['features'] : array();
```
katbailey, line 2364: Just checking that this is what's intended - should it not be
`isset($report_data['confidence']['features'][self::NO_FEATURES]) ? $report_data['confidence']['features'][self::NO_FEATURES] : array();`
?

tests/acquia_lift.test
```diff
@@ -1480,7 +1480,6 @@ class AcquiaLiftWebTestAgentAdmin extends AcquiaLiftWebTestBase {
     $this->personalizedQueue->deleteQueue();
 
     $goals = personalize_goal_load_by_conditions(array('agent' => $agent_name));
-    debug($goals);
```
katbailey, line 1483: :)

plugins/agent_types/AcquiaLiftAgent.inc
eshtadc, line 2364: See!  This is why I need your code reviews :-)

katbailey, line 2364: Oh yay - I was useful! :D

PR #110, opened 2015-01-21
CH-1498: Added start date into start campaign form.

eshtadc:


acquia_lift.admin.unibar.inc
```diff
@@ -1666,7 +1666,7 @@ function acquia_lift_campaign_start_form_submit($form, &$form_state) {
 
   // Set the end date.
   module_load_include('inc', 'personalize', 'personalize.admin');
-  personalize_agent_end_date_form_submit($agent_data->machine_name, $form_state['values']['campaign_end']);
+  personalize_agent_date_form_submit($agent_data->machine_name, $form_state['values']);
 
   // Start the campaign.
   $form_state['acquia_lift']['started'] = personalize_agent_set_status($agent_data->machine_name, PERSONALIZE_STATUS_RUNNING);
```
katbailey: This is still starting the campaign even if you select a start date in the future.

eshtadc: Oh yeah, cause you are clicking "Start campaign"... d'oh.  Let me change this so that you still have access here, but instead of "start manually" it's "start now".. and then fix the logic.  Update coming.

PR #111, opened 2015-01-23
Ch 1800: Ensure the unibar functions with older or newer versions of the navbar module

eshtadc:
Navbar module will be changing the default menu class used for all JS and CSS interaction from '.menu' to '.navbar-menu'.  This was a decision made to avoid various conflicts with the standard Drupal menu system.

See https://www.drupal.org/node/2119989


src/css/_acquia_lift.navbar.styles.scss
```diff
@@ -0,0 +1,397 @@
+/**
+ * @file
+ * Styles for integrating Personalize with the Navbar module.
+ */
+
```
Zarabadoo: I recommend adding ```@import "base";``` at the top. It will not output anything on it's own, but it will make global variables available.

src/css/_acquia_lift.navbar.styles.scss
```diff
@@ -0,0 +1,397 @@
...
+.navbar-bar .navbar-icon-acquia-lift:before {
+  background-image: url("../images/bebebe/rocket.svg");
+}
+
+.no-svg .navbar-bar .navbar-icon-acquia-lift:before {
+  background-image: url("../images/bebebe/rocket.png");
+}
+
+.navbar-bar .navbar-icon-acquia-lift {
+  &:active:before, &.navbar-active:before {
```
Zarabadoo, line 18: As a formatting rule, it would be good to put these items on separate lines. It makes it easier to parse and read.

``` scss
&:active:before, 
&.navbar-active:before
```

It can also be more condensed by formatting it as 

``` scss
&:active,
&.navbar-active {
  &:before {
    ...
  }
}
```

The same goes with similar instances.

src/css/_acquia_lift.navbar.styles.scss
```diff
@@ -0,0 +1,397 @@
...
+  }
+  .acquia-lift-menu-item {
+    margin: {
+      top: .88em;
+      bottom: .88em;
+    }
+  }
+  .navbar-level-1 {
+    .menu-wrapper > .#{$menu-class},
+    .#{$menu-class} {
```
Zarabadoo, line 77: This is awesome. And $menu-class is something that can go in _variables.scss to make it available globally.

src/css/_acquia_lift.navbar.styles.scss
eshtadc, line 18: I actually just copied this all from the original file.  Essentially what I wanted was to have all of the styles separate so that I could include them from both the regular navbar.css and navbar_1-5.css.

That's also the reason for not include base here, because it's actually included by both of those files when generated their respective css files.

src/css/_variables.scss
```diff
@@ -15,3 +15,4 @@ $pointer-size: .5em !default;
 $pointer-shadow-size: ($pointer-size + $shadow-offset) !default;
 $margin-horizontal: -1em !default;
 $winner-color: #85b200 !default;
+$menu-class: navbar-menu;
```
Zarabadoo: Add `!default` to the end of this variable.

src/css/_acquia_lift.navbar.styles.scss
Zarabadoo, line 18: I saw that once I got further down the review. I completely missed the name change and that it was a partial.

src/css/_acquia_lift.navbar.styles.scss
Zarabadoo, line 77: Which is something you have already done upon further reading.

PR #113, opened 2015-02-02
CH-1871 Refactoring the agent verification and making status change support in the unibar optional

katbailey:
The big refactoring here is moving the agent verification cache and the 'campaigns' js settings into acquia_lift module.


src/js/menu/acquia_lift.personalize.views.js
```diff
@@ -1182,10 +1182,12 @@
       if (!this.$el.hasClass('navbar-box')) {
         return;
       }
-      this.listenTo(this.collection, 'change:isActive', this.onActiveCampaignChange);
```
eshtadc, line 1185: So this approach still has events firing and listeners active, etc.

Instead I would add the following to top of the initialize:

``` javascript
if (!Drupal.settings.acquia_lift.allowStatusChange) {
  this.remove();
  return;
}
```

Then you can remove the changes in the render function because it is never called.

katbailey, line 1185: Yes, I really was fumbling around in that code :P This is much better - thanks!

PR #118, opened 2015-02-06
Issue CH-1927 by Zarabadoo: Remove gap below variations.

Zarabadoo:
- Set a minimum height of all top level menu items in the horizontal navbar.


eshtadc:
This has been merged into RC.

PR #124, opened 2015-02-27
Ch 1987 Change Lift Unibar navigation so that you can't click through to the inner page

ynx:


acquia_lift.install
```diff
@@ -203,6 +203,32 @@ function _acquia_lift_build_menu($menu_name) {
...
+  $item = array(
+    'link_title' => 'All campaigns',
+    'link_path' => 'admin/structure/personalize',
+    'menu_name' => $menu_name,
+    'options' => array(
+      'attributes' => array(
+        'aria-role' => 'button',
+        'aria-pressed' => 'false',
+        'class' => array(
+          'acquia-lift-campaign-all',
```
eshtadc: Do the individual acquia-lift-campaign-all, acquia-lift-option-sets-all, and acquia-lift-goals-all serve a particular purpose?  I don't see them referenced again.

behat-tests/README.txt
```diff
@@ -9,43 +9,44 @@ To set up your environment to run tests locally:
 1.  Download the Selenium stand-alone server JAR file:
     http://selenium-release.storage.googleapis.com/2.44/selenium-server-standalone-2.44.0.jar
 
-2.  Install Composer:
+2.  Selenium version 2.44 is only compatible with Firefox 33, but not 34 or above. You can download
```
eshtadc: I'm using Firefox 34.  The bug is specific to Firefox version 35+ I believe.  Did you try downloading the latest selenium builds and trying it out - I see talk that they are a possibility and that a new version is also coming out.  I would still reference the bug conversations as there is useful information there to help people identify problems.

behat-tests/README.txt
```diff
@@ -9,43 +9,44 @@ To set up your environment to run tests locally:
...
+    But not:
+    - Manage personalized content
+
     @todo: Add this role creation into the before/after hooks.
 
 To run tests:
 
 1.  Start the Selenium server
-    Type: java -jar /Path/to/Downloaded/selenium-server-standalone-2.44.0.jar
+    Type: java -jar /opt/selenium/selenium-server-standalone-2.44.0.jar -Dwebdriver.firefox.bin="[PATH TO YOUR Firefox.app]/Contents/MacOS/firefox-bin" &
```
eshtadc: I don't need the Dwebdriver option passed when I start Selenium.  I'm trying to find what is in my configuration for this (it's not in my path).  I'm fine with leaving it in the readme although I'd like to generalize the front part again back to /Path/to/Downloaded/ given that we don't ever mention /opt/selenium/ and people may use a different file structure.

src/js/menu/acquia_lift.personalize.views.js
```diff
@@ -943,6 +971,26 @@
    ***************************************************************/
 
   /**
+   * View for the top-level goal menu.
+   */
+  Drupal.acquiaLiftUI.MenuGoalMenuView = ViewBase.extend({
```
eshtadc: Let's pluralize this one to keep it in line with the others.  With that said I see here that two end in "MenuView" while the campaigns one doesn't... but that's outside of the scope of your ticket.

eshtadc:
Good job on the whole - you got to take a look at several parts of the module which is great!  I think you are still missing a test within behat to ensure that clicking the main headers no longer takes you to a separate page.  We want to make sure we don't break that given how confusing it has been for our users.  Perhaps we can start a new general functionality feature that isn't specific to an AB campaign or to the variations processes.  We could even make it a scenario within the access test.  Then you could pre-populate a campaign and verify the right links when it is selected and their basic functionality (limited to the scope of this issue of course).  We can always add on to it to more fully cover the unibar basics.

acquia_lift.install
ynx: The reason why "acquia-lift-campaign-all", "acquia-lift-option-sets-all", and "acquia-lift-goals-all" were created was because I see we also have "acquia-lift-campaign-new", "acquia-lift-option-sets-new", and "acquia-lift-goals-new". I wanted to keep the consistency there. 

"acquia-lift-goals-new" is used, but "acquia-lift-campaign-new" and "acquia-lift-option-sets-new" are not.

I recommend that we keep "acquia-lift-campaign-all", "acquia-lift-option-sets-all", and "acquia-lift-goals-all".

behat-tests/README.txt
eshtadc: Of course - downloading selenium versions and playing with that is also beyond the scope of this ticket.  Perhaps we could just alert Bryan.

ynx: https://selenium.googlecode.com/git/java/CHANGELOG

"Updating Native events to support Firefox 24, 31, 32 and 33"

Selenium Stand-alone server change log specifies the compatibility is up-to and including Firefox 33.  I tried with 35 and 36, they both would not work. We are able to still use Firefox 34, and for the most part it may work, but we are risking known native events incompatibility that may bite us in the future. It can be very confusing and time consuming when that happens, especially on Jenkins.

There has been talks of resolving, but I could not find a timeline. In this case I recommend that we all use Selenium 2.44.0 (starting yesterday, 2.45.0) and Firefox 33.1.1 across all dev and test environments ensure the best stability and consistency.

behat-tests/README.txt
ynx: You are right we should not have "/opt/selenium/" there. I put it there by accident. Sorry about that. I meant [DIRECTORY PATH]. It will be updated.

The reason why it is "/opt/selenium/" on my Mac local is because a lot of System Admins put Selenium there and I mimic their set up for consistency. But yes you are right, the path is arbitrary.

If you don't specify parameter webdriver.firefox.bin, I think it uses your native Firefox app PATH'ed from the system (but now you've said it, I may want to check again). I recommend that we use a copy of Firefox that is dedicated to Selenium, and specifically enforce the reference to it to keep the best stability and consistency for the test environment (on Mac local and Jenkins). This also enables us to freely update our "main" Firefox.

ynx:
Lisa, thank you very much for your encouragement! I appreciate your patience and kindness. =D

In Behat, for only testing the new three "All" links but not testing the "no-clickthrough on the main menu items" - this usually goes into a long story ...  The shortest answer that I can come up with is, the removed clickthrough behavior was undesired, it falls into the categories of bug fix or UI enhancement; the new behavior of "no-clickthrough" is not a feature, and we should only test and demonstrate features by Behat.

The longer story involves the original design purpose and usage of Behat, and risks and burdens of using Behat differently than designed for. I would love to talk to you through more interactive means to figure out what I should do in this case.

If we really have to test the "no-clickthrough" behavior, I recommend that we test in unit or functional tests.

src/js/menu/acquia_lift.personalize.views.js
ynx: Yes you are right, I should pluralize in this case.

acquia_lift.install
eshtadc: But in this case none of them (-all) are used, right?  I feel like this is just adding additional css clutter.  We don't need them for testing, styling, or semantics from what I can tell.

eshtadc:
Re Behat - I completely see your point regarding the "feature" level of this change.  I do still like the idea of having a scenario where we just test the basic functionality of the unibar when no campaigns are selected vs. a campaign of each type selected and the core links.  This is outside of the scope of this task though, but probably a backlog item.  No need to worry about it here.

acquia_lift.install
ynx: You are right, I think they are just adding useless overhead. My previous colleagues picked consistency over cleanliness, and I really don't have a strong feeling towards either (other than clarifying the reasoning). In this case, let's use the Lift team's current standard and remove the "-all" classes.

PR #125, opened 2015-02-28
CH-1969 Grey out reports link until the campaign has started.

ynx:
(for bookkeeping only)

First approach: implemented solution by JavaScript.

### DIFF
--- a/acquia_lift.module
+++ b/acquia_lift.module
@@ -3035,6 +3035,8 @@ function acquia_lift_get_campaign_details($include_status = TRUE) {
     foreach ($option_sets as $option_set) {
       $option_set_types[] = $option_set->plugin;
     }
+    $current_status = personalize_agent_get_status($agent->machine_name);
+    $report_is_active = $plugin instanceof PersonalizeAgentReportInterface && $current_status != PERSONALIZE_STATUS_NOT_STARTED;
     $campaigns[$agent->machine_name] = array(
       'name' => $agent->machine_name,
       'label' => filter_xss($agent->label),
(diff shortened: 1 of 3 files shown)


eshtadc:
I keep feeling that this should be handled in the Drupal code rather than the JavaScript code.  Whether or not reports are available for an agent could be a function of several things... with this commit we'd be splitting it up and putting half the responsibility with the PHP code and half with the backbone app.  Wouldn't it be easier to just check the status when rendering the links for settings?  https://github.com/acquia/acquia_lift/blob/develop/acquia_lift.module#L2940

ynx:
Lisa, you are right, we should do this from the back end given the chance - I simply did not find acquia_lift_get_campaign_details() and did not want to wait to ask. I have reimplemented using PHP instead of JavaScript.

eshtadc:
The base code looks great - this makes sense now.  There are some simpletest tests that need updating I believe (definitely in the "Acquia Lift Web Tests - Workflow", possibly elsewhere).  Also - I can't test manually cause the bug regarding the campaign context from this morning appears to still be in here - can you go ahead and rebase on develop?

ynx:
1) I have rebased on develop and manually tested again. The feature still work nicely.
2) All my reverting commits looked confusing, so I have squashed them into one. If you have this local checkout branch, please delete and checkout again.
3) I could not find any references to tests (other than the Behat ones). Can you please forward me a doc reference, or let me know where they are and how to run?

eshtadc:
The tests I'm referring to are the Simpletest PHP/Drupal tests.  You can find them in each module in the "tests" folder: https://github.com/acquia/acquia_lift/tree/develop/tests
You'll need to enable the "testing" module to be able to run them and then can use the site graphical interface or a drush command to run.
https://www.drupal.org/simpletest-tutorial-drupal7

ynx:
Fixed the failed test. Ready for review.

P.S. I also ran through the full Personalize test suite, and made sure all other tests are passing (except one that's on Kat in Target).

PR #126, opened 2015-03-02
Ch 1964: Created overview process to determine type of variation set for creation

eshtadc:
Alter the personalize campaign form to change the variations form.
This is limited in scope to only handling the selection portion itself.
The card.js and card.css files are from the elemental-card project: https://github.com/acquia/elemental-card
There will be more styling to come, but it should only affect the copied over css/js files and a theme function.


acquia_lift.admin.unibar.inc
```diff
@@ -594,33 +594,17 @@ function acquia_lift_option_set_add_modal_callback($ajax) {
...
   ctools_modal_add_js();
 
-  $path = drupal_get_path('module', 'acquia_lift');
+  module_load_include('inc', 'acquia_lift', 'acquia_lift.ui');
+  $types = acquia_lift_option_set_types_ui();
+  // Add a few fields specific to this display.
+  $types['personalize-element']['path'] = 'admin/structure/personalize/variations/personalize-elements/add';
+  $types['personalize-element']['modal'] = FALSE;
+  $types['personalize-block']['path'] = 'admin/structure/personalize/variations/personalize-blocks/add';
+  $types['personalize-block']['modal'] = FALSE;
```
katbailey, line 603: The keys of the array returned by the acquia_lift_option_set_types_ui() function are just 'element' and 'block'. Seems like the path at least could be included in what's returned by the function and then for the modal property you could just loop through and set them all to FALSE?

acquia_lift.admin.wizard.inc
```diff
@@ -0,0 +1,155 @@
+<?php
+/**
+ * @file personalize.admin.campaign.inc Functions specific to the Acquia Lift
```
katbailey, line 3: Wrong file name

acquia_lift.admin.unibar.inc
eshtadc, line 603: I like that - thanks.

PR #127, opened 2015-03-03
Ch 1965: Add blocks to "what" screen

eshtadc:


acquia_lift.admin.wizard.inc
```diff
@@ -147,9 +233,313 @@ function acquia_lift_personalize_campaign_wizard_targeting_alter(&$form, &$form_
...
+
+  // Common functionality needed before submitting the block submission
+  // handler whether for new or existing blocks.
+  $submitBlockHandler = function ($block_content, $option_set = NULL) use ($form, $form_state) {
+    module_load_include('inc', 'personalize_blocks', 'personalize_blocks.admin');
+    _acquia_lift_personalize_campaign_wizard_form_state_blocks_alter($form_state, $block_content, $option_set);
+    // Submit handler only runs if there is a specific button id.
+    $form_state['clicked_button']['#id'] = 'edit-submit-form';
+    personalize_blocks_form_submit($form, $form_state);
+  };
```
katbailey, line 373: Clever!

tests/acquia_lift.test
```diff
@@ -3219,32 +3219,168 @@ class AcquiaLiftWebTestCampaignWizard extends AcquiaLiftWebTestBase {
...
+    $option_set = reset($option_sets);
+    $this->assertEqual(count($option_set->options), 2);
+    $expected_queue_items[] = array(
+      'method' => 'savePoint',
+      'args' => array(
+        $agent->getMachineName(),
+        personalize_get_decision_point_name_for_option_set($option_set),
+      ),
+      'agent' => $agent->getMachineName(),
+    );
```
katbailey, line 3582: Hmm, we should have a restriction in place ensuring this new workflow can only be used with the acquia_lift_target agent type (and if you create a new agent it can only be of this type) and then there won't be any items added to the queue here. Not sure if that's an easy restriction to put in at this point?

eshtadc, line 3582: Everything up to this point applies to any agent.  I wasn't going to limit things until we got to the targeting part.  I've been doing all my testing and development with acquia_lift agents.

PR #128, opened 2015-03-03
Ch 2011 - Forms for creating nested campaigns

katbailey:


acquia_lift.admin.inc
```diff
@@ -858,3 +858,866 @@ function acquia_lift_report_conversion() {
...
+        'operator' => 'equals',
+        'match' => '',
+        'plugin' => '',
+      );
+      $num_contexts++;
+    }
+  }
+
+  $form_state['num_contexts'] = count($mappings);
+  foreach ($mappings as $delta => $mapping) {
```
eshtadc, line 984: Is all this logic duplicated?  We're going to continue to have it in both places... although I guess maybe I'll split that out once I work with it because I'll know what portions I'm re-using.

acquia_lift.admin.inc
```diff
@@ -858,3 +858,866 @@ function acquia_lift_report_conversion() {
...
+  else {
+    drupal_set_message(t('There was a problem saving the new target audience.'), 'error');
+  }
+
+}
+
+/**
+ * Saves a new target audience for an agent.
+ *
+ * @param $label
```
eshtadc, line 1163: OK - so this may be my only substantive comment, but this docblock needs to be filled out ;-)

acquia_lift.admin.inc
katbailey, line 984: Yes, this is totally duplicated code but I figured we'd be leaving the original form alone from now on anyway and it will be going away soon. The original version of this code is tied into individual options so it's probably more useful for you to work with what I've put here.

PR #129, opened 2015-03-03
CH-1817 Add a drush command to sync campaigns

ynx:
Several notable items as possible alternative approaches:
1) When selecting a campaign, we can use its "label" instead of "machine name" to make it more user-friendly.
2) We can make the "selecting a campaign" a multi-select "selecting campaigns", so the users can select more than one campaign to sync in one execution. Should be easy.
3) The command right now is user-friendly but not very shell script-friendly; so depending on the use cases, we might want to continue on the current approach, or change to just plain arguments and options instead.


acquia_lift.drush.inc
```diff
@@ -32,3 +36,43 @@ function drush_acquia_lift_menu_rebuild() {
...
+    if (!$name_selection) {
+      return;
+    }
+    $agent_name = $agent_names[$name_selection];
+    $agent = $agents[$agent_name];
+    $agents = array($agent_name => $agent);
+  }
+
+  // Step 3. Sync selected campaign(s)
+  acquia_lift_batch_sync_campaigns($agents);
```
katbailey: So my hope had been that we could actually do this in such a way that we receive feedback during batch processing, rather than seeing nothing in the console until the whole thing is complete. I think this documentation page explains how to do that: https://www.drupal.org/node/873132
It may take some refactoring though. Can you at least investigate if this is possible?

ynx: I think it is no difficulty to do as requested. The reason I didn't go for that route was because we would then not be able to reuse acquia_lift_batch_sync_campaigns(), and have to maintain two sets of code with almost the same purpose.

Let me try refactoring it using native batch.

ynx:
I have broken up the batch by agents, and slotted batch_process in between. This gives prompts for syncing each campaign. Please review and let me know if this works better.

---

P.S. Based on this code, I have also tried to further enhance the output (e.g. showing the campaign names in the success/error messages, etc), but it didn't work out. The primary reason here is because the code that I tapped into was not made for this purpose, therefore I would either need to:
- make local copies of [acquia_lift_batch_finished()](https://github.com/acquia/acquia_lift/blob/CH-1817/acquia_lift.batch.inc#L92) and [acquia_lift_batch_process_item()](https://github.com/acquia/acquia_lift/blob/CH-1817/acquia_lift.batch.inc#L52), or 
- to hack into them at where they are.

Both look very costly/messy for such small return in feature. I'll be happy to show you the code, but I really don't like how the code has went in that case.

acquia_lift.drush.inc
```diff
@@ -32,3 +36,64 @@ function drush_acquia_lift_menu_rebuild() {
...
+    }
+    $batch = array(
+      'operations' => $batch_operations,
+      'finished' => 'acquia_lift_batch_finished',
+      'file' => drupal_get_path('module', 'acquia_lift') . '/acquia_lift.batch.inc',
+    );
+    batch_set($batch);
+    $batch =& batch_get();
+    $batch['progressive'] = false;
+    drush_backend_batch_process();
```
katbailey, line 118: Hmm, I wonder if we're getting anything at all out of using the batch API here... what do you think? Maybe we should just get the list of operations for each agent, process them directly (i.e. just call the method) and output a message when each agent has had all its components sync'd?

ynx:
As per better understanding of the requirements, I have continued on using batch API but has introduced slightly more local code to make the process more responsive. The code is also now polished. Please review.

acquia_lift.drush.inc
```diff
@@ -32,3 +36,115 @@ function drush_acquia_lift_menu_rebuild() {
...
+
+  foreach ($results as $result) {
+    $message = '- ' . dt($result);
+    $message_type = 'ok';
+    if (strpos($result, ACQUIA_LIFT_OPERATION_ERROR_PREFIX) === 0) {
+      $message_type = 'error';
+    }
+    drush_log($message, $message_type);
+  }
+}
```
katbailey: OMG no new line at end of file! Yes, that is the only thing I can find to point out - this is awesome :)

ynx: Haha okay I see we needed the new line at the end of file. Added.

PR #133, opened 2015-03-07
Ch 2029: Add/edit/delete target audiences for acquia_lift_target agents.

eshtadc:


acquia_lift.admin.wizard.inc
```diff
@@ -643,3 +807,191 @@ function _acquia_lift_personalize_campaign_wizard_variations_element(&$form, &$f
...
+    '#field_suffix' => t(' of the specified contexts'),
+    '#description' => t('Choose how multiple contexts should be applied to options. Choose "any" if the rule should apply if the user has any of the contexts. Choose "all" if the rule should apply only if the user has all of the contexts.'),
+    '#options' => array(
+      'OR' => 'any',
+      'AND' => 'all'
+    ),
+    '#default_value' => $default_strategy,
+  );
+  return $form;
+}
```
katbailey, line 1014: I think there's a huge chunk of code that can now be deleted from acquia_lift.admin.inc, i.e. acquia_lift_new_target_audience_form().

acquia_lift.admin.wizard.inc
```diff
@@ -243,10 +241,68 @@ function acquia_lift_personalize_campaign_wizard_variations_alter(&$form, &$form
...
+        '#type' => 'container'
+      );
+      $form['audiences'][$rule_id]['header']['delete'] = array(
+        '#markup' => l(t('Delete'), 'admin/structure/personalize/manage/' . $agent_data->machine_name . '/audience/' . $rule_id . '/delete', array(
+          'attributes' => array(
+            'class' => array('acquia-lift-delete')
+          ),
+        )),
+      );
+      _acquia_lift_personalize_campaign_wizard_targeting_audience($form['audiences'][$rule_id]['details'], $form_state, $agent_data, $rule_id, $rule);
```
katbailey: Could we change this to 
`$form['audiences'][$rule_id]['details'] = _acquia_lift_personalize_campaign_wizard_targeting_audience($form_state, ...)`
That function already returns the form array and it seems confusing to pass the section of the form by reference here.

acquia_lift.admin.wizard.inc
```diff
@@ -444,6 +561,53 @@ function acquia_lift_personalize_campaign_wizard_variations_submit_element_edit(
...
+    $visitor_context = personalize_admin_convert_visitor_context_form_values($form_state['values']['visitor_context']);
+  }
+  $agent_data->data['visitor_context'] = $visitor_context;
+
+  // Handle targeting audiences.
+  if (empty($form_state['values']['audiences'])) {
+    return;
+  }
+  module_load_include('inc', 'acquia_lift', 'acquia_lift.admin');
+  foreach ($form_state['values']['audiences'] as $rule_id => $rule_values) {
```
katbailey: Can we call them audiences instead of rules? I think it could get confusing otherwise as an audience is made up of rules.

acquia_lift.admin.wizard.inc
```diff
@@ -643,3 +807,191 @@ function _acquia_lift_personalize_campaign_wizard_variations_element(&$form, &$f
...
+
+/**
+ * Helper function to generate an audience sub-form.
+ *
+ * @param $agent_data
+ *   The current agent data.
+ * @param $rule_parent_identifier
+ *   The identifier used for this rule, either the machine_name or "add" for
+ *   new rules.
+ * @param $rule
```
katbailey: Can haz $audience? :)

acquia_lift.admin.wizard.inc
```diff
@@ -643,3 +807,191 @@ function _acquia_lift_personalize_campaign_wizard_variations_element(&$form, &$f
...
+    '#required' => !empty($rule),
+  );
+  $form['rule_id'] = array(
+    '#type' => 'value',
+    '#value' => $rule_parent_identifier,
+  );
+
+  // This is the portion of the form that will be replace when the "add new" or "remove
+  // context" links are clicked.
+  $main_wrapper_id = 'acquia-lift-targeting';
```
katbailey: This needs to get the audience name appended to it or something - currently if you have two or more audiences you end up with multiple divs with this id and then the "Add context" button on the second audience results in a context being added to the first audience.

tests/acquia_lift.test
```diff
@@ -3770,4 +3775,141 @@ class AcquiaLiftWebTestCampaignWizard extends AcquiaLiftWebTestBase {
...
     $this->assertUrl('node');
   }
+
+  /**
+   * Verify that audience-style targeting is only available for Acquia Lift
+   * Target agents.
+   */
+  function testAudienceAvailability() {
+    $test_agent = $this->createTestAgent();
+    $target_agent = $this->createTestAgent(array('agent_type' => 'acquia_lift_target'));
```
katbailey: There's now a createTargetingAgent() method that you could use instead. The createTestAgent method is very much geared towards acquia_lift agents.

tests/acquia_lift.test
```diff
@@ -3770,4 +3775,141 @@ class AcquiaLiftWebTestCampaignWizard extends AcquiaLiftWebTestBase {
...
+    $form = array(
+      '#agent' => $agent_data,
+    );
+    acquia_lift_personalize_campaign_wizard_submit($form, $form_state);
+
+    $this->drupalGet('admin/structure/personalize/manage/' . $agent->getMachineName() . '/targeting');
+
+    // Now there should be an existing audience as well as a form to create
+    // a new audience.
+    $this->assertFieldByName('audiences[muppet-fans][details][name]', 'Muppet fans');
```
katbailey, line 3992: Yay muppets :)

acquia_lift.admin.wizard.inc
eshtadc, line 1014: I was keeping them for reference until the end of the who/what form stuff.  Do you think I should remove as I go to keep relevant things together?  I guess I can always look back in the code history.

acquia_lift.admin.wizard.inc
eshtadc: Yeah... the more I was working with these... the less sense they began to make.  There are a few functions set up this way so I would want to make them all consistent.  They are all within the wizard.inc page (adding blocks and element forms is the same way).

tests/acquia_lift.test
eshtadc: It's within a different test class - but I can pull it into the base.

tests/acquia_lift.test
eshtadc, line 3992: :-)

PR #134, opened 2015-03-09
CH-1787 Check Lift Web's dnt cookie before waiting for segments callback.

ynx:


acquia_lift_profiles/qunit/tests.js
```diff
@@ -139,7 +139,16 @@ QUnit.asyncTest( "init test", function( assert ) {
 });
 
 
+QUnit.test("Get context values with 'do not track' cookie", function( assert ) {
+  $.cookie('tc_dnt', 'true', {path:'/'});
```
katbailey, line 145: It would be good to start the test showing that without the cookie being set, we get a non-empty object returned.

acquia_lift_profiles/qunit/tests.js
```diff
@@ -139,7 +139,16 @@ QUnit.asyncTest( "init test", function( assert ) {
...
+  expect(1);
+  var contextResult = Drupal.personalize.visitor_context.acquia_lift_profiles_context.getContext({'segment1':'segment1'});
+  assert.ok($.isEmptyObject(contextResult), 'Context result must be an empty object');
+  $.cookie('tc_dnt', null, {path:'/'});
+});
+
 QUnit.asyncTest("Get context values no cache", function( assert ) {
+  $.cookie('tc_dnt', null, {path:'/'});
```
katbailey: We should just add this to the tearDown function for the module, if there is a module for these tests. If not we should add one :)

acquia_lift_profiles/qunit/tests.js
ynx, line 145: As per our discussion, we already have two other tests on non-empty object returned (just not in the same test), so it will suffice.

acquia_lift_profiles/qunit/tests.js
ynx: As per our discussion, we will move the assumptions from functions to setup/tearDown. In this case, we have moved them to setup.

PR #136, opened 2015-03-11
Ch 2028: Add audience targeting assignment

eshtadc:
Commit history is a mess... but the actual changes are correct.


acquia_lift.admin.inc
```diff
@@ -936,7 +895,7 @@ function acquia_lift_target_audience_save($label, $agent_name, $contexts, $strat
     $option_set->targeting[$machine_name] = array();
   }
   $option_set->targeting[$machine_name]['label'] = $label;
-  $option_set->targeting[$machine_name]['weight'] = count($option_set->targeting);
+  $option_set->targeting[$machine_name]['weight'] = $weight;
```
katbailey, line 898: I'm wondering what we should do if there's already an audience with that weight - should we keep incrementing / decrementing it until we get a weight that no other audience uses?

acquia_lift.admin.wizard.inc
```diff
@@ -577,34 +685,78 @@ function acquia_lift_personalize_campaign_wizard_targeting_submit(&$form, &$form
     return;
   }
   module_load_include('inc', 'acquia_lift', 'acquia_lift.admin');
+  $targeting = array();
+  $audience_count = 0;
+  // Determine the values for the "Everyone else" audience
+  $else = array(
```
katbailey: A minor point but I think this variable name can lead to confusion, because when you're reading the code and hearing it in your head you're hearing things like "else weight equals blah", which... yeah (or maybe it's just me? I read code like that :))

acquia_lift.admin.inc
katbailey, line 898: Actually never mind - I can see there's no need to worry about that here, it will be handled by the form.

eshtadc, line 898: Plus once we add in the dragons, it will be calculated automatically anyway rather than having users enter a weight.

acquia_lift.admin.wizard.inc
eshtadc: Didn't even think of that... but I'm certainly not married to it...was just lazy to type out $everyone_else.  But now that you said it I'll read it that way the rest of the time I'm working on it... so I'll change it.

PR #137, opened 2015-03-17
Ch 2308 Fix and stablize behat tests.

ynx:
1) Fixing behat tests. 
2) Removing role dependancy on 'Marketer' and 'Nonmarketer'. 
3) Add comments to organize behat steps into 'user action' blocks. 
4) Add '@insulated' tag to keep sessions and cookies discontinous between scenarios. 
5) Added the ability for the FeatureContext to reuse all contexts. 
6) Update code and readme documentation.
7) Introducing spin functions. 
8) Gearing up local context's actions and assertions with spin functions.
9) Removing 'I wait for AJAX to finish' step as much as possible from features.

### DIFF
--- a/behat-tests/features/acquia_lift_element_variations.feature
+++ b/behat-tests/features/acquia_lift_element_variations.feature
@@ -63,6 +63,7 @@ Feature: Personalize elements variations can be edited for an existing campaign.
     When I click "Add variation" in the "lift_tray" region
     Then I should see "#page-title" element in the "page_content" region is "highlighted" for editing
     And I should see the text "Edit text: <H1>" in the "dialog_variation_type_form" region
+    And the "personalize_elements_content" field should contain "Test Article Title - Updated 1"
     And I should not see the link "Edit selector" in the "dialog_variation_type_form" region
     When I fill in "Test Article Title - Updated 2" for "personalize_elements_content"
     And I click "#edit-variation-type-submit-form" element in the "dialog_variation_type_form" region
--- a/behat-tests/features/acquia_lift_element_variations.feature
+++ b/behat-tests/features/acquia_lift_element_variations.feature
@@ -1,25 +1,32 @@
-#Assumes Marketer role.
+@api @javascript @insulated @campaign
 Feature: Personalize elements variations can be edited for an existing campaign.
   In order to manage element variations in context
   As a site marketer
   I want the ability to add, edit, and delete existing personalize element variations in context.
 
-  @api @javascript @campaign
   Scenario: Add personalize elements to a campaign
+    # I have a campaign.
+    # I login with the marketer role.
+    # I am on an article page.
     Given "acquia_lift" agents:
-      | machine_name                    | label |
+      | machine_name                    | label                           |
       | testing-campaign-add-variations | Testing campaign add variations |
-    And I am logged in as a user with the "Marketer" role
+    And I am logged in as a user with the "access administration pages,access toolbar,administer visitor actions,manage personalized content" permission
+    And I am on the homepage
     And I am viewing an "Article" content:
-      | title | My article title |
-      | body  | A placeholder           |
+      | title | Test Article Title - Original |
     When I click "Acquia Lift" in the "menu" region
     And I wait for AJAX to finish
     Then I should see the link "Campaigns" in the "lift_tray" region
+
+    # I open the variation set's menu.
     When I hover over "Campaigns" in the "lift_tray" region
     And I click "Testing campaign add variations" in the "lift_tray" region
-    Then I should see the link "Variation Sets" visible in the "lift_tray" region
+    And I wait for AJAX to finish
+    Then I should visibly see the link "Variation Sets" in the "lift_tray" region
     And I should see "0" for the "variation set" count
+
+    # I bring up the "Add variation set" interface.
     When I hover over "Variation Sets" in the "lift_tray" region
     Then I should see the link "Add variation set" in the "lift_tray" region
     Then I should see the link "All variation sets" in the "lift_tray" region
@@ -31,6 +38,8 @@ Feature: Personalize elements variations can be edited for an existing campaign.
     When I click "Webpage elements" in the "modal_content" region
     And I wait for AJAX to finish
     Then I should not see the modal
+
+    # I add a new variation set.
     When I click "#page-title" element in the "page_content" region
     And I wait for AJAX to finish
     Then I should see the text "<H1>" in the "dialog_variation_type" region
@@ -38,64 +47,80 @@ Feature: Personalize elements variations can be edited for an existing campaign.
     And I wait for AJAX to finish
     Then I should not see the variation type dialog
     And I should see the text "Edit text: <H1>" in the "dialog_variation_type_form" region
-    And the "personalize_elements_content" field should contain "My article title"
-    When I fill in "Test 1" for "personalize_elements_content"
+    When I fill in "Test Article Title - Updated 1" for "personalize_elements_content"
     And I fill in "Test variation set" for "title"
-    And I press the "Save" button
+    And I click "#edit-variation-type-submit-form" element in the "dialog_variation_type_form" region
     And I wait for AJAX to finish
+
+    # I verify my variation set is created.
     Then I should see the message "The variation set has been created." in the messagebox
     When I wait for AJAX to finish
     Then I should not see the variation type form dialog
-    And I should see the text "Test 1" in the "page_content" region
+    And I should see the text "Test Article Title" in the "page_content" region
     And I should see "1" for the "variation set" count
+
+    # I bring up the "Add variation" interface.
     When I hover over "Variation Sets" in the "lift_tray" region
     Then I should see the text "Test variation set" in the "lift_tray" region
     And I should see the link "Control variation" in the "lift_tray" region
     And I should see the link "Variation #1" in the "lift_tray" region
     And I should see the link "Add variation" in the "lift_tray" region
     And I should see the link "Add variation set" in the "lift_tray" region
+
+    # I add a new variation to to the existing variation set.
     When I click "Add variation" in the "lift_tray" region
     Then "#page-title" element in the "page_content" region should have "acquia-lift-page-variation-item" class
-    And I should see the text "Edit HTML: <H1>" in the "dialog_variation_type_form" region
-    And I should not see the link "Edit selector" visible in the "dialog_variation_type_form" region
-    When I fill in "Test 2" for "personalize_elements_content"
-    And I press the "Save" button
+    And I should see the text "Edit text: <H1>" in the "dialog_variation_type_form" region
+    And I should not see the link "Edit selector" in the "dialog_variation_type_form" region
+    When I fill in "Test Article Title - Updated 2" for "personalize_elements_content"
+    And I click "#edit-variation-type-submit-form" element in the "dialog_variation_type_form" region
     And I wait for AJAX to finish
+
+    # I verify my new variation is add.
     Then I should see the message "The variation has been created." in the messagebox
     When I wait for AJAX to finish
     Then I should not see the variation type form dialog
-    And I should see the text "Test 2" in the "page_content" region
+    And I should see the text "Test Article Title - Updated 2" in the "page_content" region
     And I should see "1" for the "variation set" count
     When I hover over "Variation Sets" in the "lift_tray" region
+    Then I should see the link "Variation #1" in the "lift_tray" region
     Then I should see the link "Variation #2" in the "lift_tray" region
 
-  @api @javascript @campaign
   Scenario: Edit existing personalize elements for an acquia_lift campaign.
+    # I have a campaign and a variation set.
+    # I login with the marketer role.
+    # I am on an article page.
     Given "acquia_lift" agents:
-      | machine_name                    | label |
+      | machine_name                     | label                            |
       | testing-campaign-edit-variations | Testing campaign edit variations |
     And personalized elements:
-      | label                     | agent                            | selector     | type     | content                 |
-      | Page title updated         | testing-campaign-edit-variations | #page-title | editText | The Rainbow Connection  |
-    And I am logged in as a user with the "Marketer" role
+      | label              | agent                            | selector    | type     | content                |
+      | Page title updated | testing-campaign-edit-variations | #page-title | editText | The Rainbow Connection |
+    And I am logged in as a user with the "access administration pages,access toolbar,administer visitor actions,manage personalized content" permission
+    And I am on the homepage
     And I am viewing an "Article" content:
-      | title | My article title |
-      | body  | A placeholder           |
+      | title | Test Article Title - Original |
     When I click "Acquia Lift" in the "menu" region
     And I wait for AJAX to finish
     Then I should see the link "Campaigns" in the "lift_tray" region
+
+    # I open the variation set's menu.
     When I hover over "Campaigns" in the "lift_tray" region
     And I click "Testing campaign edit variations" in the "lift_tray" region
     And I wait for AJAX to finish
-    Then I should see the link "Variation Sets" visible in the "lift_tray" region
+    Then I should visibly see the link "Variation Sets" in the "lift_tray" region
     And I should see "1" for the "variation set" count
+
+    # I bring up the "Edit" variation set interface.
     When I hover over "Variation Sets" in the "lift_tray" region
-    Then I should see the text "Site name updated" in the "lift_tray" region
-    And I should see the link "Option A" visible in the "lift_tray" region
-    And "Site name updated" set "Control variation" variation should not have the "Edit" link
-    And "Site name updated" set "Control variation" variation should not have the "Delete" link
-    And "Site name updated" set "Option A" variation should have the "Edit" link
-    And "Site name updated" set "Option A" variation should have the "Delete" link
+    Then I should see the text "Page title updated" in the "lift_tray" region
+    And I should visibly see the link "Option A" in the "lift_tray" region
+    And "Page title updated" set "Control variation" variation should not have the "Edit" link
+    And "Page title updated" set "Control variation" variation should not have the "Delete" link
+    And "Page title updated" set "Option A" variation should have the "Edit" link
+    And "Page title updated" set "Option A" variation should have the "Delete" link
+
+    # I edit the variation set.
     When I click "Edit" link for the "Page title updated" set "Option A" variation
     And I wait for AJAX to finish
     Then I should see the text "Edit text: <H1>" in the "dialog_variation_type_form" region
@@ -103,8 +128,10 @@ Feature: Personalize elements variations can be edited for an existing campaign.
     And the "option_label" field should contain "Option A"
     When I fill in "Variation 1" for "option_label"
     And I fill in "Moving Right Along" for "personalize_elements_content"
-    And I press the "Save" button
+    And I click "#edit-variation-type-submit-form" element in the "dialog_variation_type_form" region
     And I wait for AJAX to finish
+
+    # I verify my variation is updated.
     Then I should see the message "The variation has been updated." in the messagebox
     When I wait for AJAX to finish
     Then I should not see the variation type form dialog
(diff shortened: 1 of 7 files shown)


behat-tests/features/acquia_lift_element_variations.feature
```diff
@@ -1,170 +1,198 @@
...
   Scenario: Add personalize elements to a campaign
+    # I have a campaign.
+    # I login with the marketer role.
+    # I am on an article page.
     Given "acquia_lift" agents:
-      | machine_name                    | label |
+      | machine_name                    | label                           |
       | testing-campaign-add-variations | Testing campaign add variations |
-    And I am logged in as a user with the "Marketer" role
+    And I am logged in as a user with the "access administration pages,access toolbar,administer visitor actions,manage personalized content" permission
```
eshtadc, line 14: Hmmm - thought we were going to add the code to the FeatureContext to actually create a role from a given statement rather than just showing the permissions.  The reason I hadn't used the permission list before was that I wanted the role definition to be maintained in a single place .  Maybe not so necessary for the non-marketer role, but the "marketer role" should be a bit more centrally defined.

behat-tests/features/acquia_lift_element_variations.feature
```diff
@@ -1,170 +1,198 @@
...
     And I am on the homepage
```
eshtadc: Does this mean that they are viewing article content on the home page or is the "And I am on the homepage" portion to be removed now?

behat-tests/features/acquia_lift_element_variations.feature
ynx, line 14: Ah that's the reason why you did not go for the permission approach - I was sure there was a reason. =D

The reason I was going straight to permissions, was because I thought the step was perfect for the use. This is because the plain idea the reader needs to know is "user has permissions xxx and yyy" instead of "user has a role of aaa that in term has permissions yyy, zzz". If I had went for "Given I have the role Marketer in the system", I would have had to comment "# Marketer has permissions to www,xxx,yyy,zzz but nothing else special" before each use anyway - so my (assume actual non-tech Marketers) readers would immediately know what Marketer means without reading code.

This really boils down to the fact that we are going through all the trouble implementing and maintaining this very expensive Gherkin language is only because we are now writing an non-tech readable article instead of a program, so I feel many programmer habits would have to give in to editor habits.

When we need to update the definition of "Marketer" in the future, we can always search and replace by utilizing the "marketer" comment I have already put on top of each use.

behat-tests/features/bootstrap/FeatureContext.php
```diff
@@ -197,6 +225,8 @@ public function createPersonalizedElements(TableNode $elementsTable) {
    * @Then I should see :count for the :type count
    */
   public function assertMenuCount($count, $type) {
+    $this->spinUntilAjaxIsFinished();
```
eshtadc: I understand that you are trying to move the Ajax statements out of the features (which I have mixed feelings about but can be swayed).  It feels wrong to be throwing the spin functions throughout the assertions where they aren't necessarily needed.  For example, many things are checked after AJAX finishes processing and if we throw these in throughout the assertions we're running that javascript evaluation many more times than need be.  To really do it right to remove the ajax step I feel like we'd have to write a bunch of extra custom actions that merely wrap a plain form submit/link click, etc. with a spin function as well.  At that point we are using custom functions for everything and it's out of control.

I guess - convince me why this is the best way :-)

behat-tests/features/bootstrap/FeatureContext.php
```diff
@@ -400,12 +424,14 @@ public function assertLinkVisibleRegion($link, $region) {
   }
 
   /**
-   * @Then I should not see the link :link visible in the :region( region)
+   * @Then I should not visibly see the link :link in the :region( region)
```
eshtadc, line 465: Is this change based on a standard somewhere?

behat-tests/features/acquia_lift_element_variations.feature
ynx: "logged in as a user with the permission" step would land the user on the home page, then the next "And I am viewing an 'Article' content" (from Behat Drupal Extension) creates the article node and then sends the user to the node page (I think).

To be serious, I am not a fan of how the language as defined upstream. I feel a better line would be "And I go to an article of:", at least. This would read much more clear of what's going on, which you would not have been confused by it then.

I in fact think we can enhance quite a bit of things as defined in Behat Drupal Extension haha, but that's a much larger topic, and is out of the scope for now.

behat-tests/features/bootstrap/FeatureContext.php
```diff
@@ -369,30 +402,21 @@ public function assertRegionElementHasClass($selector, $region, $class) {
   }
 
   /**
-   * @When I wait for messagebox to close
```
eshtadc, line 372: Hmm - was this really not being used anywhere?  I don't see the calls to it... but I remember the need for it was because the just checking ajax completion isn't enough to have the message box closed.  Plus sometimes we want to see something that happens before the message box closes but the drivers see ajax as still being active when it is on screen.

behat-tests/features/bootstrap/FeatureContext.php
```diff
@@ -728,4 +759,67 @@ private function getCurrentCampaign() {
...
+    }
+  }
+
+  /**
+   * Spin JavaScript evaluation.
+   *
+   * @param string  $assertionScript  Assertion script
+   * @param integer $attemptThreshold Number of attempts to execute the command.
+   */
+  private function spinJavaScriptEvaluation($assertionScript, $attemptThreshold = 15) {
```
eshtadc, line 811: I like this as a nice extendable base function :-)

behat-tests/features/acquia_lift_element_variations.feature
```diff
@@ -1,170 +1,198 @@
...
-    Then I should see the text "<SPAN>" in the "dialog_variation_type" region
+
+    # I add a new variation set.
+    When I click "#page-title" element in the "page_content" region
+    Then I should see the text "<H1>" in the "dialog_variation_type" region
     When I click "Edit text" in the "dialog_variation_type" region
-    And I wait for AJAX to finish
     Then I should not see the variation type dialog
-    And I should see the text "Edit text: <SPAN>" in the "dialog_variation_type_form" region
-    And the "personalize_elements_content" field should contain the site title
```
eshtadc, line 39: This seems to have been left out in the update.  We want to verify that the edit in context functionality pulled in the correct text.  Should be easier in this case since we aren't using something dynamic like the site title.

behat-tests/features/acquia_lift_element_variations.feature
```diff
@@ -1,170 +1,198 @@
...
-    When I fill in "Test 1" for "personalize_elements_content"
+    And I should see the text "Edit text: <H1>" in the "dialog_variation_type_form" region
+    When I fill in "Test Article Title - Updated 1" for "personalize_elements_content"
     And I fill in "Test variation set" for "title"
-    And I press the "Save" button
-    And I wait for AJAX to finish
+    And I click "#edit-variation-type-submit-form" element in the "dialog_variation_type_form" region
```
eshtadc, line 48: This seems less intuitive.  Why was it switched to using the element id rather than the button text?

behat-tests/features/acquia_lift_element_variations.feature
```diff
@@ -1,170 +1,198 @@
...
+
+    # I add a new variation to to the existing variation set.
     When I click "Add variation" in the "lift_tray" region
-    Then "#site-name a span" element in the "page_content" region should have "acquia-lift-page-variation-item" class
-    And I should see the text "Edit text: <SPAN>" in the "dialog_variation_type_form" region
-    And I should not see the link "Edit selector" visible in the "dialog_variation_type_form" region
-    When I fill in "Test 2" for "personalize_elements_content"
-    And I press the "Save" button
-    And I wait for AJAX to finish
+    Then "#page-title" element in the "page_content" region should have "acquia-lift-page-variation-item" class
```
eshtadc: Outside the scope of this... but we should probably move this logic into something more readable.   The acquia-lift-page-variation-item class indicates that the element is highlighted during edit.   Perhaps a todo to remind us next time we come to this.

behat-tests/features/acquia_lift_element_variations.feature
```diff
@@ -1,170 +1,198 @@
...
+    And I should see the text "Edit text: <H1>" in the "dialog_variation_type_form" region
+    And I should not see the link "Edit selector" in the "dialog_variation_type_form" region
```
eshtadc, line 69: Doesn't this still get seen because it is on the page but not visible?

behat-tests/features/acquia_lift_simple_ab.feature
```diff
@@ -1,146 +1,154 @@
...
+    And I should see "0" for the "variation" count
+    And I should visibly see the link "Goals" in the "lift_tray" region
+    And I should see "0" for the "goal" count
+    And menu item "Reports" should be "inactive"
+    And menu item "Start campaign" should be "inactive"
+    And I should not see the modal
+    And the variation edit mode is "active"
+
+    # I create a new variation.
+    When I click "#site-name a span" element in the "page_content" region
```
eshtadc, line 54: Doesn't this also need to be shifted to an article?  The site-name a span is part of what is problematic across environments.

behat-tests/features/bootstrap/FeatureContext.php
ynx, line 465: Yes you are right, I am following [Behat Drupal Extension](http://cgit.drupalcode.org/drupalextension/tree/src/Drupal/DrupalExtension/Context/MinkContext.php?id=857b164d18da5b0ab7208380f07c5f7620b198f7#n242). So in the future we don't have to remember which Context the steps were from.  =)

behat-tests/features/bootstrap/FeatureContext.php
ynx, line 372: It was not used. Want to keep it anyway? The piece is very small so I don't think its a huge inconvenience to carry it on anyway.

behat-tests/features/bootstrap/FeatureContext.php
ynx, line 811: Awe you are so nice. Thank you! =D

behat-tests/features/acquia_lift_element_variations.feature
ynx, line 48: Since we switched to the article node, there was another "Save" button on the page, and "clicking on 'Save'" was clicking both buttons, which messes stuff up and breaks the test.

Clicking on "#edit-variation-type-submit-form" is much uglier, you are right. I was hoping to address this with the rest of ".some-class" and "#some-id" altogether in a future release.

behat-tests/features/acquia_lift_simple_ab.feature
ynx, line 54: Yes at least for consistency, I think using article for all feature tests are better. I can switch if you feel it is needed.

#### On the other hand ...

acquia_lift_simple_ab.feature is going away in the same iteration so we can save ourselves some trouble not to do it.

Moreover, I was always hoping that we can separate the dev environment from behat test environment. i.e. test insulation in full. I have actually tried a bit, and it can be as simply done as during Behat run time:
- Create or clone over the full DB to DB_behat.
- Add another drupal connection for the DB_behat DB and switch to it.
- Tear down and repeat by @beforescenario and @insulated.

I got quite far actually, but didn't want to go pass my two-day time box for this story at the time.

If I had to choose, I'd spend time figuring out this insulation instead of using means like article node to dodge - they'll just introduce problems like the second "Save" button mentioned above.

behat-tests/features/bootstrap/FeatureContext.php
ynx: Haha I will be very happy to try to convince you on that.

#### Why "I wait for AJAX to finish" is evil

We are going through all the trouble implementing and maintaining this very expensive Gherkin language is only because we are now writing an non-tech readable article instead of a program. 

This means, at the end game of a complete Behat suite, we are going to be almost reading an series of articles that feel very much not techie, in fact, so plain that your newbie-dev, QA, PM, customer, marketer, stakeholders, and average internet surfers are able to understand what this product is doing, and they do so without calling customer service department or reading the code. And it is a live documentation so everybody can count on it.

For this reason, techie language or setup's such as _I wait for AJAX to finish_, _I wait for 7 seconds_, _I click on the link ".some-element-class"_, _I see the page response code is "200"_, _I practice some voodoo that you-don't-know_ are counter-productive to the use of Gherkin/Behat, therefore defeats the most important reason why we are using Gherkin/Behat.

Long story short, **those steps should not be seem in published feature files.**

P.S. The best reference I know on this advocation is this [talk](http://everzet.com/post/33359584167/video-of-my-talk-from-symfony-live-2012-paris) by everzet himself. I checked on him about 2 month ago and he is still saying the same thing - probably why [spin](http://docs.behat.org/en/v2.5/cookbook/using_spin_functions.html) is blessed by him. =D

#### What's the cost of spin (i.e. step prerequisite checking)?

The time cost of spin is in microseconds range.

Profiling code:

``` PHP
      echo "Start JS evaluation: " . microtime(). "\n";
      $flag = $this->getMink()->getSession()->evaluateScript($assertionScript);
      echo "End JS evaluation: " . microtime(). "\n";
      return $flag;
```

Console output:

``` Gherkin
      │ Start JS evaluation: 0.44491700 1426697195
      │ End JS evaluation: 0.48901300 1426697195
      │
    And I should not visibly see the link "Goals" in the "lift_tray" region                                                                                # FeatureContext::assertNotLinkVisibleRegion()
      │ Start JS evaluation: 0.71067300 1426697195
      │ End JS evaluation: 0.76761400 1426697195
      │
    And I should not visibly see the link "Reports" in the "lift_tray" region                                                                              # 
```

This is why wait() for 5 seconds for each JavaScript prerequisite probing is silly.

#### What's the cost of NOT spin?

When we assert something in a step, chances are we are always assuming the previous step is fully finished by all related systems. For example, _AJAX has finished_, _an element has appeared on the page after synchronous JavaScript execution_, _a DB record has already been inserted in to the database and populated to slave_, _a remote system has made an update_.

Unfortunately, because Selenium is real time, any prerequisite can be filled or it can be not filled, regardless of how big or how small the prerequisite looks. Without those checking, it is only a matter of frequency the unexpected failure halts our automated test, directly proportional to the size of uncared timeline in the full test suite.

I do realize this extensive checking introduces redundancy, the reason why I opted in, was because it is much much much more expensive to have programmers spending hours picking through all the weird failures that stops automation every couple of features. The redundant spins accumulated through hours of full suite were still way under a minute.

---

I think I'm going to stop this essay here ... But yes I realize this is an open topic across Behat world so I am definitely open to discussion on better approaches.

behat-tests/features/acquia_lift_element_variations.feature
eshtadc, line 14: But the non-technical reader thinks in terms of roles.  I wouldn't even remember the right list of permissions.  Only Drupal programmers think in terms of specific permissions.

behat-tests/features/acquia_lift_element_variations.feature
eshtadc: But do we need both is the question?  I don't _think_ we need the "And I am on the homepage" anymore.

behat-tests/features/acquia_lift_element_variations.feature
ynx, line 14: Permission is not set up by a programmer, but by a Drupal site admin in site's UI. This site admin is not necessarily a programmer, and is often not. This means Permission is a really Drupal feature, and we are really saying "before using the Lift feature, the user needs a set-up in the Permission feature".

We can always introduce Roles as the collection of permissions, but then the argument falls back to the fact that we will still have to somehow clarify at each use that the Role "marketer" is a vehicle for a list of four specific permissions and nothing more, so we are not really cutting any maintenance here.

behat-tests/features/acquia_lift_element_variations.feature
ynx: I see what you are saying now. Yes you are right, I agree in this case we should remove "And I am on the homepage" because this assertion serves no purpose.

behat-tests/features/acquia_lift_element_variations.feature
ynx, line 39: You are right, this was left out. The problem at the time was the the field actually contains not only the title, but a bunch of white spaces and newlines before and after the title, and there was no way to assert that using the existing tool set.

screenshot

---

I have a question on this one. I feel the Lift JavaScript needs to be enhanced. The text() as picked up from the page title element needs to be trim()'ed before entering into the modal textarea. 

And then we can use a normal step:

``` Gerkin
﻿And the "personalize_elements_content" field should contain "Test Article Title - Original"
```

Please let me know if you agree this is the case.

behat-tests/features/acquia_lift_element_variations.feature
ynx: Yes, what we need here is a new step like:

``` Gherkin
Then "#page-title" element in the "page_content" region is highlighted for editing
```

We can do that in this story if you think proper. Should be simple.

behat-tests/features/acquia_lift_element_variations.feature
ynx, line 69: If I remember correctly, we can no longer find the link by `$regionObj->findLink($link);`, therefore the "visible" step fails but the current one does not.

behat-tests/features/acquia_lift_element_variations.feature
ynx: Implemented.

behat-tests/features/acquia_lift_element_variations.feature
ynx: Implemented.

behat-tests/features/bootstrap/FeatureContext.php
eshtadc, line 465: That's not what I meant.... sorry let me try again.
The step is one that is defined in our own FeatureContext.  The question is why the change to the syntax?  Trying to find if there is a standard of syntax perhaps that would have the "visibly" before "see the link" rather than "see the link.... visible".

behat-tests/features/bootstrap/FeatureContext.php
eshtadc, line 372: Nope - if it's not being used then get rid of it :-)

behat-tests/features/acquia_lift_element_variations.feature
eshtadc, line 39: I don't think we necessarily want to change the functionality to trim the text.  I do remember seeing this case now that you mention it... but we want to allow users to start with the same text that is on their site or within their custom option.  We don't want users making inadvertent changes due to our "helpfulness".  Can we check to see if the element _contains_ the text then rather than is equivalent?  It at least lets us know if it is completely broken if not a perfect check.

behat-tests/features/acquia_lift_element_variations.feature
eshtadc, line 48: That makes sense.  Can you add a todo comment so we remember why this is the case?  Right now it looks more like its just to call the spin function that you added so it is good to know there is a larger reason.

behat-tests/features/acquia_lift_simple_ab.feature
eshtadc, line 54: Fair enough... you are removing it next.  Just keep an eye when you do to make sure any remaining functionality is covered in the elements/goals tests (at least as much as it was covered in the AB test).

behat-tests/features/bootstrap/FeatureContext.php
ynx, line 465: There is no Behat standard for defining these step language, so "visibly" or "visible" would have been both good. However, we do have a similar existing step in our imported Context, Drupal MinkContext, `@Then I should not visibly see the link :link`.

The benefit of aligning the language is, we now have:

``` Gherkin
# from MinkContext.php
@Then I should not visibly see the link :link
# from FeatureContext.php
@Then I should not visibly see the link :link in the :region( region)
```

behat-tests/features/bootstrap/FeatureContext.php
ynx, line 372: Then can I also delete iWaitSeconds()? This is a popular step, but the only good use of it was in debugging, and we already have a very nice `@Then I break` from DrupalContext.php.

behat-tests/features/bootstrap/FeatureContext.php
eshtadc, line 465: Gotcha - that's what I was curious about.  All for being standard across the contexts.  Thanks for cleaning this up.

behat-tests/features/bootstrap/FeatureContext.php
eshtadc, line 372: Yup - noticed that in there when I was looking at your code.  I don't even know why its there cause I just use "Then I break" when I'm debugging (or just run in debug mode).

behat-tests/features/acquia_lift_element_variations.feature
ynx, line 39: Yes I have had the same reasoning as yours, that's why this one has been so hesitant for me. The only reason I noted it is because HTML ignores the leading and trailing space chars anyway, and those space chars are frequently added for DOM indentation, so I thought trimming them was actually bettering user experience.

But ya I think there are two sides here and I can't tell which one is better.

I couldn't find a "contain" support, only "equivalent". But let me try again, worst I will just have to write a new one.

behat-tests/features/acquia_lift_element_variations.feature
ynx, line 48: Implemented.

behat-tests/features/bootstrap/FeatureContext.php
ynx, line 372: Cool. Deleted.

behat-tests/features/bootstrap/FeatureContext.php
ynx, line 465: Awesome!

behat-tests/features/acquia_lift_simple_ab.feature
ynx, line 54: Sure!

behat-tests/features/acquia_lift_element_variations.feature
ynx, line 39: Implemented.

ynx:
Already squash-merged into develop.

PR #139, opened 2015-03-20
Ch 2030: Dragon Drop

eshtadc:
The history looks weird from merging in workflow.... but at least now this is just based on workflow :-)


acquia_lift.admin.wizard.inc
```diff
@@ -299,6 +296,20 @@ function acquia_lift_personalize_campaign_wizard_targeting_alter(&$form, &$form_
     $variation_options[$option['option_id']] = $option['option_label'];
   }
 
+  // The "Everyone else" audience should be anything explicitly assigned plus
+  // any options that aren't assigned.
```
katbailey: Per our discussion, the "everyone else" audience needs to be created up front and all variations assigned to it - this would happen when the variation set is created.

acquia_lift.admin.wizard.inc
```diff
@@ -248,9 +248,6 @@ function acquia_lift_personalize_campaign_wizard_targeting_alter(&$form, &$form_
   // Acquia Lift Target agents allow for audience designation to be used across
   // the agent.
   $agent_data = $form['#agent'];
-  if ($agent_data->plugin !== 'acquia_lift_target') {
-    return;
-  }
```
katbailey, line 253: Was this change intended?

eshtadc, line 253: Yes because the entire alteration doesn't get run now unless it is an acquia_lift_target campaign so this would be a duplicate check.  Since its the only agent that will be editable from Acquia Lift anyway it seemed easiest.

katbailey, line 253: Ah ok cool

acquia_lift.admin.inc
```diff
@@ -1241,3 +1242,58 @@ function acquia_lift_get_option_set_for_targeting($agent_name) {
...
+    }
+  }
+  $targeting_options = array();
+  foreach ($option_set->options as $option) {
+    $targeting_options[] = $option['option_id'];
+  }
+  // Any unassigned option is automatically assigned to the "Everyone else"
+  // category.
+  $unassigned_options = array_diff($targeting_options, $assigned);
+  acquia_lift_save_targeting_structure($agent, array(ACQUIA_LIFT_TARGETING_EVERYONE_ELSE => $unassigned_options));
```
katbailey, line 1290: I may not be reading this correctly but this acquia_lift_save_targeting_structure() call will only result in the `$agent->data['lift_targeting']` property being populated, right? So on subsequent calls to acquia_lift_validate_minimum_targeting() the `if (empty($option_set->targeting))` check will still evaluate to true and it'll save the "everyone else" audience again. I realize that probably won't actually do anything, just checking it's what's intended here. Like maybe you should check the `$agent->data['lift_targeting']` property as well?

acquia_lift.admin.wizard.inc
```diff
@@ -299,6 +303,20 @@ function acquia_lift_personalize_campaign_wizard_targeting_alter(&$form, &$form_
...
+  foreach ($saved_targeting as $audience_id => $options) {
+    $assigned_options = array_merge($assigned_options, $options);
+  }
+  $unassigned_options = array_diff_key($variation_options, array_flip($assigned_options));
+  if (isset($saved_targeting[ACQUIA_LIFT_TARGETING_EVERYONE_ELSE])) {
+    $saved_targeting[ACQUIA_LIFT_TARGETING_EVERYONE_ELSE] += array_keys($unassigned_options);
+  }
+  else {
+    $saved_targeting[ACQUIA_LIFT_TARGETING_EVERYONE_ELSE] = array_keys($unassigned_options);
+  }
```
katbailey: If more options are being added to the everyone else audience, will it be clear to the user that hitting the "next" button will result in a change that will then need to be reviewed?

katbailey: Ack, I think I'm just confused - I don't understand why the unassigned options are being added here and also below in the unassigned audience. It will probably make sense when I test it out ;)

acquia_lift.admin.wizard.inc
```diff
@@ -386,15 +414,34 @@ function acquia_lift_personalize_campaign_wizard_targeting_alter(&$form, &$form_
...
-  $form['audiences']['add']['details'] = _acquia_lift_personalize_campaign_wizard_targeting_audience($form, $form_state, $agent_data, 'add');
+  // If the agent has been started, then there is the possibility of unassigned
+  // variation options that need to be assigned to a region in order to be
+  // available for assignment via drag and drop.
+  if ($status !== PERSONALIZE_STATUS_NOT_STARTED) {
+    $assigned_options = array();
+    foreach ($saved_targeting as $audience_id => $options) {
+      $assigned_options = array_merge($assigned_options, $options);
+    }
+    $unassigned_options = array_diff_key($variation_options, array_flip($assigned_options));
```
katbailey, line 411: Why does this need to be evaluated again?

tests/acquia_lift.test
```diff
@@ -3666,67 +3682,20 @@ class AcquiaLiftWebTestCampaignWizard extends AcquiaLiftWebTestBase {
...
-          $agent->getTitle(),
-          'adaptive',
-          PERSONALIZE_STATUS_NOT_STARTED,
-          10/100,
-          20/100,
-          TRUE,
-        ),
-        'agent' => $agent->getMachineName(),
-    );
-    $this->assertQueueItems($expected_queue_items);
```
katbailey, line 3725: A small taste of what's coming soon... to a PR near you... :)

acquia_lift.admin.inc
eshtadc, line 1290: You're right - this is backwards.  It may only really need to check the agent property rather than the option set property because by the time that is populated, you are starting the agent and shouldn't be automatically changing things anyway.

acquia_lift.admin.wizard.inc
eshtadc, line 411: This is the display of unassigned options that get shown when variations are added to an option set after the agent is started.  Since they are no longer automatically assigned to everyone else - then they need to appear in their own special "unassigned" area so that they are still in the UI to drag to an audience.

acquia_lift.admin.wizard.inc
eshtadc: The idea here is that unassigned options are being added to the everyone else audience automatically before the agent is started.  This allows them to be included, for example, if the customer never even goes to targeting (because they want a simple test).  Once the campaign has been started, nothing happens automatically anymore (hence the need for the unassigned options area).

acquia_lift.admin.inc
eshtadc, line 1290: Although... the audience definition is still saved to the option set.  So that part is correct.  It's just that the targeting is saved on the agent.

katbailey, line 1290: Oh right, so $option_set->targeting won't be empty on subsequent calls. It makes sense as it is then, right?

eshtadc, line 1290: But its a valid point - the check for unassigned options should happen against the agent property and not option set like I have it here.   Posting the change in a moment - just re-ran the tests.

acquia_lift.admin.wizard.inc
katbailey: OK but it looks like the above piece of code is being executed regardless of campaign status?

PR #140, opened 2015-03-24
Ch 2189: Implement goals screen

eshtadc:


plugins/agent_types/AcquiaLiftTarget.inc
```diff
@@ -61,4 +61,20 @@ class AcquiaLiftTarget extends PersonalizeAgentBase implements PersonalizeExplic
...
+  public function useClientSideGoalDelivery() {
+    return TRUE;
+  }
+
+  /**
+   * Implements PersonalizeAgentGoalInterface::sendGoal().
+   */
+  public function sendGoal($goal_name, $value = NULL) {
+    return;
+  }
```
katbailey, line 79: Heh, this is one of the changes I made in my branch and then realized you would already have made it. It also made me realize that these functions are totally unused, but I'll refrain from ripping out that interface too ;)

eshtadc, line 79: Well - I didn't actually do anything with it the implementation.  I only implemented it because I was using it to determine if we should show goals.  That's how we used it in unibar too (and possibly some other places).

PR #141, opened 2015-03-24
Ch 2304 As an engineer I want the AB campaign code removed so that the codebase remains maintainable

ynx:
Removed functional support (i.e. PHP and JavaScript) for this story. Work is still in progress for updating css and tests.

Please review only, we should not merge yet.


acquia_lift.module
```diff
@@ -2853,35 +2778,16 @@ function acquia_lift_command_messagebox($message, $seconds = 0) {
...
 /**
- * Returns an AJAX command to toggle the page variation element selector.
+ * Returns an AJAX command to toggle the element variation element selector.
  *
  * @param bool $start
- *   True to start the page variation mode, FALSE to stop it.
+ *   True to start the element variation mode, FALSE to stop it.
  */
-function acquia_lift_command_page_variation_toggle($start = TRUE) {
+function acquia_lift_command_element_variation_toggle($start = TRUE) {
```
eshtadc: I would make this one acquia_lift_command_variation_toggle($start = TRUE).  That way it is lined up with the JS command that it calls.  The only reason we had both before was because they both called the acquia_lift_command_variation_toggle JS command and specified the type.

src/js/flow/acquia_lift.variations.views.js
```diff
@@ -45,7 +45,7 @@
        *   True if showing the highlight, false if no highlight should be shown.
        */
       highlightAnchor: function(show) {
-        var highlightClass = 'acquia-lift-page-variation-item';
```
eshtadc, line 48: You'll find this in your next round of work but this class is referred to by the qunit page_variations.tests.js.

src/js/flow/acquia_lift.variations.views.js
```diff
@@ -386,7 +381,7 @@
      */
     VariationTypeMenuListView: Backbone.View.extend({
       tagName: 'ul',
-      className: 'acquia-lift-page-variation-list',
+      className: 'acquia-lift-element-variation-list',
```
eshtadc, line 384: More in the qunit tests to watch out for.  However..... those tests may be obsolete now due to the Behat tests.  They never were great but they were all we had before Behat.

src/js/menu/acquia_lift.personalize.collections.js
```diff
@@ -76,7 +76,7 @@
     },
 
     /**
```
eshtadc: I _think_ that the variations and the resetVariations functions are now obsolete as well.  They were convenience functions for dealing with the page variations structure.

src/js/menu/acquia_lift.personalize.behaviors.js
```diff
@@ -36,18 +36,13 @@
...
           }
         });
 
-        // Clear the variations for all page variation campaigns.
-        ui.collections.campaigns.each(function (model) {
-          if (model instanceof Drupal.acquiaLiftUI.MenuCampaignABModel) {
-            model.get('optionSets').resetVariations();
-          }
-        });
+        // Clear the variations for all element variation campaigns.
```
eshtadc: This is not clearing but simply merging the option sets from Drupal settings into the Backbone app (and not specific to elements).

src/js/menu/acquia_lift.personalize.behaviors.js
```diff
@@ -240,7 +228,7 @@
                   switch (type) {
                     case 'option_sets':
                       // If the menu already has a link for this setting, abort.
-                      if (!$menu.find('[data-acquia-lift-personalize-agent="' + obj.agent + '"][data-acquia-lift-personalize-id="' + key + '"].acquia-lift-preview-page-variation').length) {
+                      if (!$menu.find('[data-acquia-lift-personalize-agent="' + obj.agent + '"][data-acquia-lift-personalize-id="' + key + '"].acquia-lift-preview-element-variation').length) {
```
eshtadc: There is no equivalent usage now of acquia-lift-preview-element-variation.  All of the usages of the changed class were removed.  As a result - I think this class should just be removed rather than updated.

src/js/menu/acquia_lift.personalize.views.js
```diff
@@ -934,37 +635,19 @@
...
-          return obj.original_index == variation_index;
-        });
-
-        if (!variation) return;
-        var i, num = variation.options.length, current;
-        // Run the executor for each option in the variation.
-        for (i=0; i < num; i++) {
-          current = variation.options[i];
-          Drupal.personalize.executors[current.executor].execute($(current.selector), current.option.option_id, current.osid);
+      // Standard tests just call the executors on the selected option.
```
eshtadc: This line is no longer valid since "standard" tests compared to what?  I think leaving the second two lines can still be helpful to remember what the model is representing in the callback.

src/js/menu/acquia_lift.personalize.views.js
```diff
@@ -1436,11 +1119,6 @@
       this.listenTo(this.model, 'change:isActive', this.onEditModeChange);
       this.listenTo(this.campaignCollection, 'change:isActive', this.onCampaignChange);
 
-      this.onPageVariationEditModeProxy = $.proxy(this.onPageVariationEditMode, this);
```
eshtadc, line 1439: I think you want to keep this one around.  The event is still being triggered and we do want to have the menu react by ending the edit mode when it is triggered.

eshtadc, line 1439: Feel free to rename though.

acquia_lift.module
ynx: Agreed and implemented.

src/js/flow/acquia_lift.variations.views.js
ynx, line 48: Yes, the Qunit tests have showed up in my search for this one. To avoid extra work, I have recorded and delayed all test-related changes.

src/js/flow/acquia_lift.variations.views.js
ynx, line 384: Yes, the Qunit tests have showed up in my search for this one. To avoid extra work, I have recorded and delayed all test-related changes.

We'll see to removing or updating them at fixing tests stage.

src/js/menu/acquia_lift.personalize.collections.js
ynx: By "variations", you meant getVariations, right?

1) I can confirm getVariations() is no longer referenced, so I am removing it now.
2) resetVariations() still have one use in the triggerChange() just above, although removing it doesn't look like it breaks anything. I don't know how to check this for sure, though ... But, removing.

src/css/_acquia_lift.navbar.styles.scss
```diff
@@ -291,7 +291,7 @@
 /* Additional edit links are only available in horizontal view */
 #navbar-administration {
   .navbar-tray-vertical {
-    .acquia-lift-campaign-edit, .acquia-lift-variation-delete, .acquia-lift-variation-rename, .acquia-lift-variation-edit, .acquia-lift-goal-rename, .acquia-lift-goal-delete, .acquia-lift-page-variation-toggle {
+    .acquia-lift-campaign-edit, .acquia-lift-variation-delete, .acquia-lift-variation-rename, .acquia-lift-variation-edit, .acquia-lift-goal-rename, .acquia-lift-goal-delete, .acquia-lift-element-variation-toggle {
```
eshtadc: The toggle was removed so this class can just be removed.

css/acquia_lift.personalize.theme.css
```diff
@@ -1,9 +1,9 @@
-/* Page variation edit toggle in unified navigation bar */
```
eshtadc, line 1: The toggle is removed so this whole section that was affected here can be removed.

behat-tests/features/bootstrap/FeatureContext.php
```diff
@@ -368,18 +368,18 @@ public function assertVariationEditMode($expected_state) {
     if (!in_array($expected_state, array('active','inactive','hidden','disabled'))) {
```
eshtadc: assertVariationEditMode goes away because there is no toggle anymore.

src/js/menu/acquia_lift.personalize.behaviors.js
ynx: Agreed. Updated to:

``` PHP
// Merging settings' option_sets into campaigns' option_sets.
```

acquia_lift.admin.unibar.inc
```diff
@@ -302,7 +80,7 @@ function acquia_lift_element_variation_details_form($form, &$form_state, $variat
   $form['agent'] = array(
     '#type' => 'hidden',
   );
-  // Variation number applicable to page variations.
+  // Variation number applicable to element variations.
```
eshtadc: Looks like this is only used as a value and never populated by the client-side so there is no reason to have it here as a hidden form value.  Also no need to give it a comment when there is only one type of variation being created/edited.

acquia_lift.module
```diff
@@ -2853,35 +2778,16 @@ function acquia_lift_command_messagebox($message, $seconds = 0) {
...
-
-/**
- * Returns an AJAX command to preview a specific page variation.
- *
- * @param string $agent_name
- *   The name of the campaign/agent to which the variation belongs.
- * @param int $variation_index
- *   The index of the variation to preview.
- */
-function acquia_lift_command_page_variation_preview($agent_name, $variation_index) {
```
eshtadc, line 2917: You actually call this function from acquia_lift.unibar.inc with the name "acquia_lift_command_element_variation_preview" so it seems that this is still necessary but with a name change.

src/js/menu/acquia_lift.personalize.behaviors.js
ynx: Removing the whole `case 'option_sets':`, and changing the switch 'campaign' statement to an if.

acquia_lift.admin.unibar.inc
```diff
@@ -981,7 +479,7 @@ function acquia_lift_element_variation_delete_form($form, &$form_state, $osid, $
 }
 
 /**
- * Submit handler for acquia_lift_page_variation_delete_form().
+ * Submit handler for acquia_lift_element_variation_delete_form().
```
eshtadc, line 473: :-)
Love that copy pasta

src/js/menu/acquia_lift.personalize.views.js
ynx: Okay. Deleting `// Standard tests just call the executors on the selected option.`.

src/js/flow/acquia_lift.variations.js
```diff
@@ -53,18 +53,15 @@
   };
```
eshtadc, line 53: Also need to look at Drupal.behaviors.acquiaLiftVariations in this file.  It reads either "element" or "page" from toolbarEditMode.  This is set in the acquia_lift.module function acquia_lift_init from the $_SESSION['acquia_lift_element_trigger'] value which is in turn set by acquia_lifty.admin.wizard.inc acquia_lift_personalize_campaign_wizard_variations_submit_element_add function.  Since you no longer check the type and it won't matter anymore... for now it could just be a boolean or an empty check to set whether or not to trigger variation editing.  In the future, this could be the option_id to edit when we get to that point.

src/js/menu/acquia_lift.personalize.views.js
ynx, line 1439: How is this event being triggered? I couldn't find it any more and I am still having trouble doing so haha.

src/css/_acquia_lift.navbar.styles.scss
ynx: Okay. Removing `, .acquia-lift-element-variation-toggle`.

ynx: Does this mean the following classes are removed as well?
```
acquia-lift-element-variation-toggle-hidden
acquia-lift-element-variation-toggle-disabled
acquia-lift-element-variation-toggle-active
```

behat-tests/features/bootstrap/FeatureContext.php
ynx: Yep. This is deleted along with the simple_ab feature, and to be pushed with the rest of Behat updates.

css/acquia_lift.personalize.theme.css
ynx, line 1: Okay. Removed until the next section "QTips". This also means the image _edit-toggle-disabled.png_ is no longer referenced. Is this expected? If so, should I remove this image?

acquia_lift.admin.unibar.inc
ynx: Okay. Removing the whole:

``` PHP
  // Variation number applicable to element variations.
  $form['variation_number'] = array(
    '#type' => 'hidden',
  );
```

acquia_lift.admin.unibar.inc
ynx, line 473: I am actually quite amazed the amount of copy pasta that I am _not_ seeing in this code base. You are Kat are pretty neat. Gotta step up my game now haha.

acquia_lift.module
```diff
@@ -565,7 +533,7 @@ function acquia_lift_target_access($agent) {
  * Implements hook_personalize_agent_presave().
  */
 function acquia_lift_personalize_agent_presave($agent) {
-  if (!in_array($agent->plugin, array('acquia_lift', 'acquia_lift_simple_ab'))) {
```
eshtadc, line 584: Let's go ahead and convert this to use acquia_lift_is_testing_agent().  Will be clearer for the next round of changes.

acquia_lift.module
eshtadc, line 2917: I think this is why I can't create an element variation from the front-end.  I get a 500 error returned based on the missing function.

ynx, line 2917: It is very strange that:
1) I was still able to create/edit/delete a web element variation with no error at all.
2) acquia_lift_command_page_variation_preview traces only to acquia_lift.personalize.command.js acquia_lift_page_variation_preview(), and acquia_lift_page_variation_preview works only on pageVariations object, which pageVariations is no longer being used anywhere else. 

I think there is definitely something fixable here, I'm just not sure if bring those code back is the good fix. I'll look into this more.

acquia_lift.module
ynx, line 584: Agreed. Done.

src/js/flow/acquia_lift.variations.js
ynx, line 53: I can implement this change but I would rather leave it be. This is because making this change does not seem to produce any significant clean up benefit, and is only to be updated again in the near future.

If you still think this is necessary, then sure I can do. I am thinking on top of it, we rename the session variable to `$_SESSION['acquia_lift_edit_mode']` to be more descriptive - so at least we'll get something nice out of it.

How do I produce this trigger? I couldn't find any "start edit mode" that's sending me over to another page, seems like this "wizard" really hides himself well ...

src/js/menu/acquia_lift.personalize.views.js
eshtadc, line 1439: https://github.com/acquia/acquia_lift/blob/CH-2304/src/js/flow/acquia_lift.variations.js#L74

css/acquia_lift.personalize.theme.css
eshtadc, line 1: Go for it.

acquia_lift.module
eshtadc, line 2917: We should still have the ability to indicate that we want to preview a specific variation... so yes - this functionality should stay.  I'm not sure why you aren't getting errors.  I see what you mean about the old variation preview JS command... that is definitely not what we keep.  The answer may lie here: https://github.com/acquia/acquia_lift/blob/CH-2304/src/js/menu/acquia_lift.personalize.commands.js#L6

src/js/flow/acquia_lift.variations.js
eshtadc, line 53: One of the things about agile (at least to me) is what we are planning to do in the near future... may not happen when we expect or at all.  I'd like to make this a simple boolean since it makes no sense to indicate the trigger type.  It is still semantically accurate as a boolean.  You access this by choosing to create an element variation (in the what) screen.  After you enter the URL, this submit function kicks in to set the variable that Drupal can read upon reload to initialize the DOM selector.

src/js/menu/acquia_lift.personalize.views.js
ynx, line 1439: That's the code that triggers it, I meant the user action to trigger it and show an unexpected behavior. Nevertheless, I am able to see some errors that you are seeing now, so I am bringing those code back as suggested.

css/acquia_lift.personalize.theme.css
ynx, line 1: Done.

acquia_lift.module
ynx, line 2917: So it turns out as soon as I apply my feature's update hook (which seem to have nothing to do with the menu, the following link stops returning me a json but the default html "structure" page response:
http://localhost:8888/acquia_lift/admin/structure/acquia_lift/variation/editText

I am still debugging.

ynx, line 2917: Fixed. The reason for this error is that I accidentally removed this router configuration. The config was brought back in 5582632a416998ab394f41da6afdeee728a87200, so it works now.

This also explains why I didn't get the error until I applied my update hook - only by hook execution was the menu_router table's content regenerated, therefore taking out the routing record for _admin/structure/acquia_lift/variation/editText_ and releasing the link to default html action.

Lisa, to fix your local site for further reviews, the hook needed to be redone with the fixed code.

src/js/flow/acquia_lift.variations.js
ynx, line 53: Okay. [Done](https://github.com/acquia/acquia_lift/commit/dd989f542e11b064896fb7394454eb050cee70dd).

acquia_lift.module
eshtadc, line 2917: Yeah - when you run update.php it clears the cache.  So the menu routing table is re-read at that time.  You just hadn't cleared your cache until that point so it still showed.

ynx, line 2917: Ya it's always the cache.

On the other hand, I read the code and think the function acquia_lift_command_element_variation_preview (originally acquia_lift_command_page_variation_preview) and its if was switching by campaign type, so I have [removed it](https://github.com/acquia/acquia_lift/commit/a5cb54eb0c75ba62cbb2f67ab684c7dc32ee6ffb).

PR #142, opened 2015-03-27
Ch 2289 - various changes to support the move over to acquia_lift_target as the primary agent type

katbailey:


plugins/agent_types/AcquiaLiftAgent.inc
```diff
@@ -532,20 +477,6 @@ class AcquiaLiftAgent extends PersonalizeAgentBase implements PersonalizeAgentGo
   }
 
   /**
-   * Implements PersonalizeExplicitTargetingInterface::explicitTargetingSupportMultiple().
-   */
-  public static function explicitTargetingSupportMultiple() {
-    return PersonalizeExplicitTargetingInterface::EXPLICIT_TARGETING_MULTIPLE_BOTH;
-  }
-
-  /**
```
eshtadc, line 352: Is it on purpose that this agent doesn't override supportsAutoTargeting()?

acquia_lift.admin.unibar.inc
```diff
@@ -219,8 +216,6 @@ function acquia_lift_campaign_type_create_form_submit($form, &$form_state) {
   $agent_data->data['cache_decisions'] = $form_state['values']['cache_decisions'];
   $agent = personalize_agent_save($agent_data);
   personalize_set_campaign_context($agent->machine_name);
-  // @todo find out how to remove this in favor of asynchronous processing.
```
eshtadc: Guess you found out how to remove it ;-)

acquia_lift.admin.wizard.inc
```diff
@@ -16,7 +16,10 @@
  */
 function acquia_lift_personalize_campaign_wizard_process_bar_alter(&$form, &$form_state, $form_id) {
   $agent = $form['#agent'];
-
+  $nested = acquia_lift_get_nested_tests($agent);
```
eshtadc, line 19: From a usability point of view I'm a little wary of this one.  It's weird to users when sometimes they have options and sometimes they don't.  I think an alternative would be to take them to the place where they can confirm their changes upon click.  Of course - I guess that doesn't exist yet.

acquia_lift.batch.inc
```diff
@@ -10,11 +10,12 @@
  * Batch syncs all campaigns to Lift.
  */
 function acquia_lift_batch_sync_campaigns($agents = array()) {
-  // First delete the queue because we are about to sync everything now.
-  acquia_lift_delete_queue();
-  // Now get all sync operations for all existing Acquia Lift agents.
+  // Now get all sync operations for all existing Acquia Lift testing agents.
   if (empty($agents)) {
     foreach (acquia_lift_get_agent_types() as $type => $info) {
+      if ($type == 'acquia_lift_target') {
```
eshtadc, line 16: Was there a reason to not use acquia_lift_is_testing_agent() or did you just think the function call overhead wasn't worth it?

acquia_lift.drush.inc
```diff
@@ -48,6 +48,9 @@ function drush_acquia_lift_campaign_sync() {
   // Step 1. Exit if there is no campaign to be synchronized.
   $agents = array();
   foreach (acquia_lift_get_agent_types() as $type => $info) {
+    if ($type == 'acquia_lift_target') {
```
lkhaas: Same question about acquia_lift_is_testing_agent().

acquia_lift.install
```diff
@@ -639,6 +638,186 @@ function acquia_lift_update_7017() {
...
+      'option_count' => count($options),
+      'options' => $options,
+      'targeting' => unserialize($row->targeting),
+    );
+  }
+  // If an agent has more than one option set or if its option set contains fixed
+  // targeting rules, it cannot be converted and will be treated as a legacy agent.
+  $legacy_agents = $convertibles = array();
+  // If personalize module is not enabled all campaigns will be treated as legacy.
+  $personalize_enabled = module_exists('personalize');
```
eshtadc, line 712: This can be run with personalize not enabled?  Is this one of those weird site factory things?

acquia_lift.install
```diff
@@ -639,6 +638,186 @@ function acquia_lift_update_7017() {
...
+        $main_os->decision_name = $option_set_details['decision_name'];
+        $main_os->targeting = array(
+          ACQUIA_LIFT_TARGETING_EVERYONE_ELSE => array(
+            'osid' => $nested_osid
+          )
+        );
+        personalize_option_set_save($main_os);
+        // Convert the first of the original option sets to a nested option set
+        // and delete the others.
+        if ($i == 0) {
```
eshtadc, line 811: I'm missing why the rest can be deleted :-(

acquia_lift.install
```diff
@@ -639,6 +638,186 @@ function acquia_lift_update_7017() {
...
 /**
+ * Miscellaneous changes required for the new workflow, includes converting legacy
+ * campaigns to "personalizations".
+ */
+function acquia_lift_update_7018() {
+  // The queue is now gone so delete the queue class variable.
+  variable_del('queue_class_acquia_lift_sync');
+
+  // Get all old-style agents and convert them where possible.
+  $result = db_query("SELECT a.machine_name, a.label, o.plugin, o.osid, o.decision_name, o.options, o.targeting FROM {personalize_agent} a INNER JOIN {personalize_option_sets} o ON o.agent = a.machine_name WHERE a.plugin IN ('acquia_lift', 'acquia_lift_simple_ab')");
```
eshtadc: I thought you were only doing acquia_lift agents... I think Shay was looking into the AB agents but I'm sure he'd be happy to know he doesn't have to.

acquia_lift.module
```diff
@@ -40,27 +40,6 @@ function acquia_lift_menu() {
     'type' => MENU_CALLBACK,
     'file' => 'acquia_lift.ui.inc',
   );
-  $items['acquia_lift/personalize_in_context/start'] = array(
```
eshtadc, line 43: I was just seeing those and thinking I should do a side ticket for the extra junk that was in there.  Glad you got it :-)

acquia_lift.module
```diff
@@ -222,11 +201,29 @@ function acquia_lift_menu() {
...
   return $items;
 }
 
 /**
+ * Implements hook_menu_alter().
+ */
+function acquia_lift_menu_alter(&$items) {
+  // Change the access callback for the edit campaign page so that we can control
+  // access to Lift campaigns.
+  $items['admin/structure/personalize/manage/%personalize_agent/edit']['access callback'] = 'acquia_lift_campaign_edit_access';
```
eshtadc, line 186: I'll just grab all these when I get to that part of the clean-up.

acquia_lift.module
```diff
@@ -578,6 +526,27 @@ function acquia_lift_target_access($agent) {
...
+  $legacy_campaigns = variable_get('acquia_lift_legacy_agents', array());
+  return user_access('manage personalized content') &&
+  !empty($legacy_campaigns);
+}
+
+/**
+ * Access callback for campaign edit page.
+ */
+function acquia_lift_campaign_edit_access($agent) {
+  if (in_array($agent->plugin, array('acquia_lift', 'acquia_lift_simple_ab'))) {
```
lkhaas: I still like acquia_lift_is_testing_agent()

plugins/agent_types/AcquiaLiftAgent.inc
katbailey, line 352: Heh, well there kind of is still an open question as to whether we should just rip out support for auto-targeting now, but I didn't mean to go ahead and do that just yet so this is an oversight. Thanks for catching! :)

acquia_lift.admin.unibar.inc
katbailey: :)

acquia_lift.module
```diff
@@ -618,23 +587,6 @@ function acquia_lift_personalize_agent_status_presave($agent_name, $status) {
 }
 
 /**
- * Implements hook_personalize_agent_update_status().
- */
-function acquia_lift_personalize_agent_update_status($agent_name, $old_status, $status) {
-  // Clear the agent's verification status.
-  acquia_lift_agent_clear_verified_status($agent_name);
```
eshtadc: Does this step still need to happen (verified status cache clear)?

acquia_lift.admin.wizard.inc
katbailey, line 19: Yeah, I actually think maybe the sync with Lift button can just go away now that syncing happens at very specific times using batch api. As to reset data, that should probably only show up on the report for the test?

acquia_lift.module
```diff
@@ -687,25 +640,6 @@ function acquia_lift_personalize_visitor_context() {
 }
 
 /**
- * Implements hook_personalize_agent_delete().
- */
-function acquia_lift_personalize_agent_delete($agent_data) {
-  if (!acquia_lift_is_testing_agent($agent_data)) {
-    return;
-  }
-  acquia_lift_agent_clear_verified_status($agent_data->machine_name);
```
eshtadc: Same thing here... I don't see the verification status stuff being fully removed.

acquia_lift.module
```diff
@@ -813,191 +707,24 @@ function acquia_lift_personalize_option_set_save($option_set) {
...
-    ));
-    // Mark this agent as "pending" as it will be started once the queue runs.
-    acquia_lift_set_pending_status($agent->machine_name, $agent->label);
-  }
 }
 
 /**
  * Implements hook_personalize_option_set_delete().
  */
 function acquia_lift_personalize_option_set_delete($option_set) {
```
eshtadc: Was there a reason for leaving the empty function behind?

acquia_lift.module
```diff
@@ -813,191 +707,24 @@ function acquia_lift_personalize_option_set_save($option_set) {
...
-  acquia_lift_sync_goals($agent, $old_goals, $new_goals, TRUE);
-  // This agent will need to be re-verified as the changes may have changed
-  // whether or not it is valid.
-  acquia_lift_agent_clear_verified_status($goal_info['agent']);
 }
 
 /**
  * Implements hook_personalize_goal_delete().
  */
 function acquia_lift_personalize_goal_delete($goal_info) {
```
eshtadc: Another empty function.

acquia_lift.batch.inc
katbailey, line 16: I was actually going to use that and then saw it takes an actual agent, not a type as its param. I think once simple a/b is gone we can simplify all these checks anyway.

acquia_lift.install
katbailey, line 712: Yep, update functions get run for disabled modules too. Nothing to do with Site Factory, that's just how we first encountered the problem, because there were SF sites that had tried out Lift, created some campaigns, then disabled the modules.

acquia_lift.install
katbailey, line 811: Because for each one we've already created a new option set on the parent agent. Each of them will reference the same nested test in their targeting property.

acquia_lift.module
```diff
@@ -3175,3 +2756,56 @@ function acquia_lift_agent_clear_verified_status($agent_name = NULL) {
...
+  $agents = array();
+  foreach ($option_set->targeting as $audience_name => $audience) {
+    if (isset($audience['osid'])) {
+      if ($option_set = personalize_option_set_load($audience['osid'])) {
+        $agents[] = $option_set->agent;
+      }
+    }
+  }
+  return $agents;
+}
```
eshtadc: I found something!  Ok.... well the big red arrow in github found something.  Line break.

acquia_lift.admin.wizard.inc
eshtadc, line 19: Yeah - those both make sense to me.

acquia_lift.batch.inc
eshtadc, line 16: Gotcha.

acquia_lift.install
```diff
@@ -639,6 +638,186 @@ function acquia_lift_update_7017() {
 }
 
 /**
+ * Miscellaneous changes required for the new workflow, includes converting legacy
+ * campaigns to "personalizations".
+ */
+function acquia_lift_update_7018() {
```
eshtadc: You and Shay have the same update number in your update hooks.... something to watch out for in merging.

acquia_lift.install
```diff
@@ -639,6 +638,186 @@ function acquia_lift_update_7017() {
...
+    drupal_set_message(t('Personalize module was disabled so existing campaigns could not be converted to the new style personalizations. They will be treated as legacy campaigns.'), 'warning');
+    variable_set('acquia_lift_legacy_agents', $legacy_agents);
+    return;
+  }
+
+  // Now convert anything that can be converted.
+  foreach ($convertibles as $original_agent_name => $details) {
+    // Create a new acquia_lift_target agent which will delegate to this one.
+    $agent = new stdClass();
+    $agent->label = 'Parent personalization for ' . $details[0]['agent_label'];
```
eshtadc: This ends up really long and repetitive in the unibar.... maybe just "Converted: " or something simpler?

plugins/agent_types/AcquiaLiftAgent.inc
```diff
@@ -1024,97 +901,6 @@ class AcquiaLiftAgent extends PersonalizeAgentBase implements PersonalizeAgentGo
...
-   * Returns the operations needed to sync goals to Acquia Lift.
-   *
-   * @param $option_sets
-   *   An array representing the option sets whose targeting rules need to be sync'd.
-   * @return array
-   *    An array of items representing API calls to be made to Acquia Lift.
-   */
-  public function getFixedTargetingSyncOperations($option_sets) {
```
eshtadc, line 855: This function is still called from the batch process.
Fatal error: Call to undefined method AcquiaLiftAgent::getFixedTargetingSyncOperations() in /Users/lisa.backer/Documents/workspace/lift/acquia_lift/acquia_lift.batch.inc on line 196

acquia_lift.module
katbailey: Oh right, what I was thinking was that we'd just entirely remove the ability to change agent status from the unibar, but I forgot to actually do that part :/

acquia_lift.module
katbailey: I was thinking we might need to do something but then never got around to it, will remove it for now.

acquia_lift.module
katbailey: Yep, will remove - if there's anything that needs to be done here I can tackle it in the other story

PR #146, opened 2015-03-31
Ch 2280: Clean-up

eshtadc:
- Clean up card display
- UI visual clean-up
- Clean up targeting links
- Terminology updates
- Audience assignment changes


acquia_lift.module
```diff
@@ -3078,6 +3077,7 @@ function acquia_lift_get_campaign_details($include_status = TRUE) {
       continue;
     }
     $supports_goals = $plugin instanceof PersonalizeAgentGoalInterface;
+    $supports_targeting = $plugin instanceof PersonalizeAutoTargetingInterface || $plugin instanceof PersonalizeExplicitTargetingInterface;
```
katbailey, line 3080: These checks about what the plugin supports seem odd to me here now. I guess it's still technically possible for a personalize_target campaign to be included here? Maybe we should only support acquia_lift_target campaigns here and then we don't need to check interfaces?

katbailey, line 3080: Anyway, that's probably beyond the scope here.

PR #149, opened 2015-04-08
Ch 2394 As a Lift user I want the default Identity Type param to be configurable so that my default can be something other than email

ynx:
Please provide feedback especially on the description.


acquia_lift_profiles/acquia_lift_profiles.admin.inc
```diff
@@ -54,16 +54,25 @@ function acquia_lift_profiles_admin_form($form, &$form_state) {
     '#description' => t('Check this if you want Acquia Lift Profiles to capture the identity of the user upon login or registration. This means sending their email address to the Acquia Lift Profile Manager.'),
     '#default_value' => variable_get('acquia_lift_profiles_capture_identity', FALSE),
   );
+  $identity_parameter_display_value = variable_get('acquia_lift_profiles_identity_param') ?: 'identity';
+  $identity_type_parameter_display_value = variable_get('acquia_lift_profiles_identity_type_param') ?: 'identityType';
+  $default_identity_type_display_value = variable_get('acquia_lift_profiles_default_identity_type') ?: ACQUIA_LIFT_PROFILES_DEFAULT_IDENTITY_TYPE_DEFAULT;
```
katbailey, line 59: Ooh, for some reason I thought the ?: shorthand required a later version of PHP but I just checked and it works in 5.3 so we're good :)

acquia_lift_profiles/acquia_lift_profiles.module
```diff
@@ -6,6 +6,7 @@
  */
 
 define('ACQUIA_LIFT_PROFILES_THUMBNAIL_WIDGET_SEPARATOR', '|');
+define('ACQUIA_LIFT_PROFILES_DEFAULT_IDENTITY_TYPE_DEFAULT', 'email');
```
katbailey, line 9: Why 'DEFAULT' twice?

acquia_lift_profiles/acquia_lift_profiles.admin.inc
ynx, line 59: Haha yes 5.3. It is a good assist to variable_get(), since variable_get() defaults from null but not empty string.

acquia_lift_profiles/acquia_lift_profiles.module
ynx, line 9: This was intentional. Like THUMBNAIL_WIDGET_SEPARATOR is "thumbnail widget's separator value" above, DEFAULT_IDENTITY_TYPE_DEFAULT is "default identity type's default value". "Default identity type" itself wouldn't be a constant.

ynx:
Merged into develop.

PR #151, opened 2015-04-09
Ch 2458: Add personalized fields display to "What" subform.

eshtadc:


acquia_lift.admin.wizard.inc
```diff
@@ -923,15 +932,19 @@ function acquia_lift_personalize_campaign_wizard_variations_submit(&$form, &$for
...
-  unset($form_state['redirect']);
-  // Set the clicked button ID back.
-  $form_state['clicked_button']['#id'] = $click_id;
+
+  if ($requires_cleanup) {
+    // Clear out any redirects set by option set forms.
+    unset($form_state['redirect']);
+    // Set the clicked button ID back.
+    $form_state['clicked_button']['#id'] = $click_id;
+  }
```
katbailey, line 947: I realize this logic was added previously and here we're just restricting when it is used, but is this because we're calling personalize_blocks_form_submit() to creating / updating a personalized block? Seems like we could use some better abstraction over in personalize_blocks module, like have a function called personalize_blocks_save_personalized_block() similar to what we have in personalize_fields module. I think this might make things cleaner.
This is beyond the scope of this PR, just want to get your thoughts :)

eshtadc, line 947: Yeah - I agree.  Each time I see this code it bugs me just a little bit more.  I can fit it in this sprint as a follow up in the workflow branches.

PR #153, opened 2015-04-13
Ch 2297

katbailey:


acquia_lift.admin.inc
```diff
@@ -851,20 +851,64 @@ function acquia_lift_target_audience_save($label, $agent_name, $contexts, $strat
...
+
+  return $tests_to_delete;
+}
+
+/**
  * Takes whatever is in the 'lift_targeting' data property and converts it into
  * the required campaign structure (including nested tests where needed).
  *
  * @param stdClass $agent
  *   The agent to create the targeting structure for.
```
eshtadc, line 886: Missing the updated return parameter for deleted test agents.

acquia_lift.admin.inc
```diff
@@ -938,15 +997,13 @@ function acquia_lift_update_targeting($agent) {
  *   The name of hte targeting audience to assign the option to.
  * @throws \PersonalizeException
  */
-function acquia_lift_set_option_for_audience($agent, $option_id, $audience) {
-  $option_set = acquia_lift_get_option_set_for_targeting($agent->machine_name);
+function acquia_lift_set_option_for_audience(&$option_set, $option_id, $audience) {
```
eshtadc, line 1009: Docblock updates for all of these.....
I feel like I'm going to earn some sort of docblock nag title :-)

acquia_lift.admin.wizard.inc
```diff
@@ -71,7 +71,32 @@ function acquia_lift_personalize_campaign_wizard_variations_alter(&$form, &$form
...
+    $mvt_name = acquia_lift_get_mvt_name_for_personalization($agent_data->machine_name);
+    if ($mvt = personalize_agent_load($mvt_name)) {
+      $option_sets = personalize_option_set_load_by_agent($mvt_name);
+    }
+  }
+
+  $form['variations']['variation_set_handling'] = array(
+    '#type' => 'radios',
+    '#options' => array(
+      ACQUIA_LIFT_DECISION_LOCKSTEP => t('Lock step'),
```
eshtadc, line 90: It will probably come up at mid-sprint review, but just in case this is really just a reminder to both of us to bring up that these probably need some sort of brief description assigned to them.

acquia_lift.admin.wizard.inc
```diff
@@ -582,21 +625,49 @@ function acquia_lift_personalize_campaign_wizard_review_alter(&$form, &$form_sta
 
     // Add audiences to the summary section.
   $agent_data = $form['#agent'];
+  if (isset($agent_data->data['variation_set_handling']) && $agent_data->data['variation_set_handling'] == ACQUIA_LIFT_DECISION_MULTIVARIATE) {
+    $form['review']['summary_column']['summary']['option_sets'] = array(
+      '#markup' => '<div class="personalize-wizard-summary-item clearfix"><span class="personalize-wizard-summary-count">1</span><div class="personalize-wizard-summary-details"><div class="item-list"><h3>Multivariate Test</h3></div></div></div>'
```
eshtadc: This should use the personalize_wizard_summary_count theme function for the markup.   Since there isn't a list - it can just do it the same way that the audiences do further below in this same function.

acquia_lift.admin.wizard.inc
```diff
@@ -582,21 +625,49 @@ function acquia_lift_personalize_campaign_wizard_review_alter(&$form, &$form_sta
...
+      $num_options = count($first_os->options);
+      foreach ($option_sets as $os) {
+        if (count($os->options) != $num_options);
+        $equal_option_counts = FALSE;
+      }
+    }
+    // Prevent submission if the numbers of options aren't the same.
+    if (!$equal_option_counts) {
+      // @todo Get proper text for this.
+      drupal_set_message(t('You have multiple variation sets with different numbers of variations - the personalization cannot be started until all variation sets have the same number of variations.'), 'warning');
```
eshtadc: We should show this message in the "Let's fix this" section - https://github.com/acquia/personalize_module/blob/workflow/personalize.admin.campaign.inc#L1158

acquia_lift.admin.wizard.inc
```diff
@@ -605,6 +676,28 @@ function acquia_lift_personalize_campaign_wizard_review_alter(&$form, &$form_sta
     $form['review']['warnings_column']['warnings']['#type'] = 'container';
     $form['review']['warnings_column']['warnings']['#theme'] = 'acquia_lift_card';
   }
+  $form['#submit'] = array(
```
eshtadc: This is going to be one of those merge things... this submit isn't needed anymore because the review subform implements the standard submit callback now in the workflow branch.
https://github.com/acquia/acquia_lift/blob/workflow/acquia_lift.admin.wizard.inc#L610

acquia_lift.admin.wizard.inc
```diff
@@ -605,6 +676,28 @@ function acquia_lift_personalize_campaign_wizard_review_alter(&$form, &$form_sta
...
+ */
+function acquia_lift_personalize_campaign_wizard_review_submit($form, &$form_state) {
+  $agent_data = $form['#agent'];
+  module_load_include('inc', 'acquia_lift', 'acquia_lift.admin');
+  module_load_include('inc', 'acquia_lift', 'acquia_lift.batch');
+
+  try {
+    $agent_data->tests_to_delete = acquia_lift_implement_test_structure($agent_data);
+    acquia_lift_batch_sync_tests(array($agent_data->machine_name => $agent_data));
+    // If the "Start" button was pressed then start the campaign.
```
eshtadc: It's possible to save this form on the review page without hitting the start button.  Should be checked like we do in the base personalize module: https://github.com/acquia/personalize_module/blob/workflow/personalize.admin.campaign.inc#L1391

acquia_lift.admin.wizard.inc
```diff
@@ -840,6 +933,9 @@ function acquia_lift_personalize_campaign_wizard_submit(&$form, &$form_state) {
     $submit_function($form, $form_state, $agent_data, $agent_instance);
   }
 
+  if (isset($form_state['values']['variations']['variation_set_handling'])) {
+    $agent_data->data['variation_set_handling'] = $form_state['values']['variations']['variation_set_handling'];
+  }
```
eshtadc: I think this makes more sense in acquia_lift_personalize_campaign_wizard_variations_submit unless there is a reason it should be in the base submit handler?

acquia_lift.batch.inc
```diff
@@ -119,20 +116,27 @@ function acquia_lift_batch_finished($success, $results, $operations) {
...
- *
- * @param $agents
- *   An array of agents with agent names as keys and stdClass objects rep-
- *   resenting the agents as values.
- * @return array
- *   An array of operation arrays, each with a 'method' key and an 'args'
- *   key, that can be processed by acquia_lift_sync_item().
- */
-function acquia_lift_get_sync_operations_for_agents($agents) {
+function acquia_lift_get_sync_operations_for_personalization($personalization) {
```
eshtadc: Docblock sync-up... but I think this is the first time we're introducing "personalization" as a code term rather than just in text.  I think we should probably set some expectation around the words agent, campaign, and personalization within the code.  I worry it will soon be at that point when someone asks what a function is and we have to give them the history of the product in order to explain it.

acquia_lift.module
```diff
@@ -643,6 +647,21 @@ function acquia_lift_personalize_option_set_save($option_set) {
 }
 
 /**
+ * Implements hook_personalize_option_set_delete().
+ */
+function acquia_lift_personalize_option_set_delete($option_set) {
+  $agent = personalize_agent_load($option_set->agent);
```
eshtadc, line 664: Does the agent need to be checked here to make sure its a Lift agent?

plugins/agent_types/AcquiaLiftTarget.inc
```diff
@@ -53,14 +53,16 @@ class AcquiaLiftTarget extends PersonalizeAgentBase implements PersonalizeExplic
...
   }
 
   /**
    * Implements PersonalizeAgentInterface::supportsMultipleDecisionPoints().
    */
   public function supportsMultipleDecisionPoints() {
-    return FALSE;
+    // We don't really, but at the time of option set creation, decision point
+    // names cannot be set.
+    return TRUE;
```
eshtadc, line 65: So the advanced settings on the variations subform shouldn't show the "decision name" field anymore, right?

acquia_lift.admin.wizard.inc
eshtadc, line 90: Again - for mid-sprint review.  Maybe this could at least not be shown until there are more than one variation sets added.  Perhaps even showing it on the add variation set card and then putting it under the advanced options for the agent after that?  Seems awfully prominent right now when it may not even apply.

acquia_lift.admin.wizard.inc
```diff
@@ -582,21 +625,49 @@ function acquia_lift_personalize_campaign_wizard_review_alter(&$form, &$form_sta
...
+  else {
+    // If there are multiple option sets, check that they all have the same number
+    // of options.
+    $option_sets = personalize_option_set_load_by_agent($agent_data->machine_name);
+    $equal_option_counts = TRUE;
+    if (count($option_sets) > 1) {
+      $first_os = array_shift($option_sets);
+      $num_options = count($first_os->options);
+      foreach ($option_sets as $os) {
+        if (count($os->options) != $num_options);
```
eshtadc: The ";" here is causing the $equal_option_counts to always be set to FALSE.

acquia_lift.admin.wizard.inc
katbailey: Ha! Pardon my _extreme_ laziness there :P

acquia_lift.admin.wizard.inc
katbailey: Doh!

acquia_lift.admin.wizard.inc
katbailey: I think my only reason was that this is where changes to the agent are being saved. Otherwise we need to call personalize_agent_save() again in the variations submit handler.

acquia_lift.batch.inc
katbailey: Yep, good point - let's just stick with agent in code ;)

acquia_lift.module
katbailey, line 664: Yes :)

plugins/agent_types/AcquiaLiftTarget.inc
katbailey, line 65: Right, I had made a note of that in my todo list - but if it's something you get to before I do feel free to remove it.

acquia_lift.admin.wizard.inc
eshtadc: Since its the same agent instance getting passed to the steps, we can change things in the steps and then at the very end the save gets called in the main submit without having to call it over and over in each step.  It's how a few other changes already are handled - like the stuff in the review form and the agent advanced settings, etc.

katbailey: Ah ok, perfect - will move it then.

PR #154, opened 2015-04-14
Ch 1770 Change Vocabulary Mappings to use any visitor context; make vocabulary a visitor context too

ynx:
I am creating this pull request so Kat can review my current approach. I feel I may not have implemented in the most informed ways, that's why I'm hoping we can take a look right now and figure out how I can do better.

### DIFF
--- a/acquia_lift_profiles/acquia_lift_profiles.admin.inc
+++ b/acquia_lift_profiles/acquia_lift_profiles.admin.inc
@@ -106,9 +106,9 @@ function acquia_lift_profiles_admin_form($form, &$form_state) {
   module_load_include('inc', 'personalize', 'personalize.admin');
   $groups = personalize_get_grouped_context_options(NULL, TRUE, array('acquia_lift_profiles_context', 'acquia_lift_context'));
 
-  // Vocabulary mappings.
-  $form['acquia_lift_profiles_vocabulary_mappings'] = array(
-    '#title' => t('Vocabulary Mappings'),
+  // Field mappings.
+  $form['acquia_lift_profiles_field_mappings'] = array(
+    '#title' => t('Field Mappings'),
     '#description' => t('For the "content section", "content keywords", and "persona" fields in Acquia Lift Profiles, you can map them to different visitor contexts.'),
     '#type' => 'fieldset',
     '#tree' => TRUE,
@@ -118,12 +118,12 @@ function acquia_lift_profiles_admin_form($form, &$form_state) {
     'content_keywords' => 'Content Keywords',
     'persona' => 'Persona',
   );
-  $vocabulary_mappings = variable_get('acquia_lift_profiles_vocabulary_mappings', array());
+  $field_mappings = variable_get('acquia_lift_profiles_field_mappings', array());
   foreach ($field_list as $field_name => $field_friendly_name) {
-    $default_value = isset($vocabulary_mappings[$field_name]) ? $vocabulary_mappings[$field_name] : NULL;
+    $default_value = isset($field_mappings[$field_name]) ? $field_mappings[$field_name] : NULL;
     $select = acquia_lift_profiles_admin_build_visitor_context_select($groups, $default_value);
-    $form['acquia_lift_profiles_vocabulary_mappings'][$field_name] = $select;
-    $form['acquia_lift_profiles_vocabulary_mappings'][$field_name]['#title'] = t($field_friendly_name);
+    $form['acquia_lift_profiles_field_mappings'][$field_name] = $select;
+    $form['acquia_lift_profiles_field_mappings'][$field_name]['#title'] = t($field_friendly_name);
   }
 
   // Only show the "Tracked actions" selector if acquia_lift_profiles has already been
@@ -223,9 +223,9 @@ function acquia_lift_profiles_admin_form_submit($form, &$form_state) {
   foreach ($variables as $key) {
     variable_set($key, $form_state['values'][$key]);
   }
-  // Vocabulary mappings. Don't save empty entries.
-  $acquia_lift_profiles_vocabulary_mappings = array_filter($form_state['values']['acquia_lift_profiles_vocabulary_mappings']);
-  variable_set('acquia_lift_profiles_vocabulary_mappings', $acquia_lift_profiles_vocabulary_mappings);
+  // Field mappings. Don't save empty entries.
+  $acquia_lift_profiles_field_mappings = array_filter($form_state['values']['acquia_lift_profiles_field_mappings']);
+  variable_set('acquia_lift_profiles_field_mappings', $acquia_lift_profiles_field_mappings);
   // Identity parameters.
   $identity_variables = array(
     'acquia_lift_profiles_capture_identity',
--- a/acquia_lift_profiles/acquia_lift_profiles.admin.inc
+++ b/acquia_lift_profiles/acquia_lift_profiles.admin.inc
@@ -101,43 +101,37 @@ function acquia_lift_profiles_admin_form($form, &$form_state) {
     )),
     '#default_value' => variable_get('acquia_lift_profiles_default_identity_type', ''),
   );
-  $vocabularies = taxonomy_get_vocabularies();
-  $mappings = variable_get('acquia_lift_profiles_vocabulary_mappings', array());
-  $vocab_options = array('' => t('Select...'));
-  foreach ($vocabularies as $vid => $voc) {
-    $vocab_options[$vid] = $voc->name;
-  }
+
+  // Retrieving all context options.
+  module_load_include('inc', 'personalize', 'personalize.admin');
+  $groups = personalize_get_grouped_context_options(NULL, TRUE, array('acquia_lift_profiles_context', 'acquia_lift_context'));
+
+  // Vocabulary mappings.
   $form['acquia_lift_profiles_vocabulary_mappings'] = array(
     '#title' => t('Vocabulary Mappings'),
+    '#description' => t('For each vocabularies available in Acquia Lift Profiles, you can map a visitor context, whose value will then be sent as the value for that vocabulary.'),
     '#type' => 'fieldset',
     '#tree' => TRUE,
   );
-  $form['acquia_lift_profiles_vocabulary_mappings']['content_section'] = array(
-    '#title' => t('Content Section'),
-    '#description' => t('The vocabulary to use for the content section'),
-    '#type' => 'select',
-    '#options' => $vocab_options,
-    '#default_value' => isset($mappings['content_section']) ? $mappings['content_section'] : '',
+  $type_list = array(
+    'content_section',
+    'content_keywords',
+    'persona',
   );
-  $form['acquia_lift_profiles_vocabulary_mappings']['content_keywords'] = array(
-    '#title' => t('Content Keywords'),
-    '#description' => t('The vocabulary to use for the content keywords'),
-    '#type' => 'select',
-    '#options' => $vocab_options,
-    '#default_value' => isset($mappings['content_keywords']) ? $mappings['content_keywords'] : '',
-  );
-  $form['acquia_lift_profiles_vocabulary_mappings']['persona'] = array(
-    '#title' => t('Persona'),
-    '#description' => t('The vocabulary to use for the persona'),
-    '#type' => 'select',
-    '#options' => $vocab_options,
-    '#default_value' => isset($mappings['persona']) ? $mappings['persona'] : '',
-  );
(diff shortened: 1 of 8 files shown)


acquia_lift_profiles/acquia_lift_profiles.admin.inc
```diff
@@ -165,7 +158,7 @@ function acquia_lift_profiles_admin_form($form, &$form_state) {
         if ($i == $min) {
           $form['acquia_lift_profiles_udf_mappings'][$type][$udf]['#prefix'] = '<div class="acquia-lift-profiles-hidden-udfs">';
         }
-        elseif ($i == $max - 1) {
+        if ($i == $max - 1) {
```
ynx, line 165: Changing this "elseif" to "if" is fixing a bug not related to the story. The "elseif" is logically incorrect and introduces misconfiguration on the "show all" button at the edge case when $min is 3 and $max is 4.

acquia_lift_profiles/acquia_lift_profiles.install
```diff
@@ -180,3 +180,19 @@ function acquia_lift_profiles_update_7002(&$sandbox) {
...
+  // Migration: prefix to existing mappings (if not already).
+  // Before: array ('', '7', '82', 'taxonomy_context__4')
+  // After: array ('', 'taxonomy_context__7', 'taxonomy_context__82', 'taxonomy_context__4')
+  $vocabulary_mappings_updated = preg_replace('/^(\d+)$/', 'taxonomy_context' . preg_quote(PERSONALIZE_TARGETING_ADMIN_SEPARATOR) . '$1', $vocabulary_mappings);
+  // If taxonomy is already configured, enable the taxonomy context module.
+  if ($vocabulary_mappings != $vocabulary_mappings_updated) {
+    module_enable(array('personalize_taxonomy_context'));
+  }
+  variable_set('acquia_lift_profiles_vocabulary_mappings', $vocabulary_mappings_updated);
+}
```
ynx, line 220: @katbailey You at one point noted "you'd also need to make acquia_lift_profiles module declare a dependency on personalize_taxonomy". I now think this may actually not necessarily the case. Pfizer/Tim's use case may just be his custom context module, and no taxonomy. Because using taxonomy as the vocabulary mapping is voluntary, I did not declare acquia_lift_profiles' dependencies on taxonomy_context and taxonomy, and left the choice open to the user.

I understand that I may be mistaken the use cases here, so please let me know haha.

Two things at migration:
1) Update the saved value by prefixing the vocabulary id.
2) Only enable the personalize_taxonomy_context if it is already in use before this update hook.

acquia_lift_profiles/acquia_lift_profiles.module
```diff
@@ -459,17 +463,27 @@ function acquia_lift_profiles_node_page_context($node = NULL, $primary = FALSE)
...
+            '@module_name' => $module_name,
+            '@function_name' => $get_terms_function_name,
+          ));
+          throw new AcquiaLiftProfilesFunctionNotExistException($error_message);
         }
+        $terms = call_user_func($get_terms_function_name, $node->nid, $context);
+        // Make sure page context terms are unique.
+        $context_terms = array_merge($page_context[$name], $terms);
+        $page_context[$name] = array_unique($context_terms);
       }
```
ynx: I am really self-criticizing on this design here. Please advise your thoughts.

katbailey: See my comment in the other PR - we shouldn't need to do anything on the PHP side now that these are visitor contexts. In our js we'll grab the context values and set them as the relevant properties on the page context object: https://github.com/acquia/acquia_lift/blob/7.x-1.x/acquia_lift_profiles/js/acquia_lift_profiles.js#L241

acquia_lift_profiles/acquia_lift_profiles.install
katbailey, line 220: It definitely needs to be a dependency as this is part of the core functionality of acquia_lift_profiles and people won't know anything about personalize_taxonomy_context module.

ynx, line 220: Good to know. [Done](71db33b3e38e12573c4ce85cd60be259c16b17de).

acquia_lift_profiles/acquia_lift_profiles.admin.inc
```diff
@@ -101,43 +101,37 @@ function acquia_lift_profiles_admin_form($form, &$form_state) {
...
-  }
+
+  // Retrieving all context options.
+  module_load_include('inc', 'personalize', 'personalize.admin');
+  $groups = personalize_get_grouped_context_options(NULL, TRUE, array('acquia_lift_profiles_context', 'acquia_lift_context'));
+
+  // Vocabulary mappings.
   $form['acquia_lift_profiles_vocabulary_mappings'] = array(
     '#title' => t('Vocabulary Mappings'),
+    '#description' => t('For each vocabularies available in Acquia Lift Profiles, you can map a visitor context, whose value will then be sent as the value for that vocabulary.'),
```
katbailey: They are not vocabularies in Acquia Lift Profiles. Let's say something like "For the 'site section', 'content keywords' and 'persona' fields in Acquia Lift Profiles, you can map a visitor context..."

acquia_lift_profiles/acquia_lift_profiles.admin.inc
```diff
@@ -101,43 +101,37 @@ function acquia_lift_profiles_admin_form($form, &$form_state) {
...
+    'content_section',
+    'content_keywords',
+    'persona',
   );
+  $vocabulary_mappings = variable_get('acquia_lift_profiles_vocabulary_mappings', array());
+  foreach ($type_list as $type) {
+    $default_value = isset($vocabulary_mappings[$type]) ? $vocabulary_mappings[$type] : NULL;
+    $select = acquia_lift_profiles_admin_build_visitor_context_select($groups, $default_value);
+    $type_name = str_replace('_', ' ', $type);
+    $type_title = ucfirst($type_name);
```
katbailey: Let's make $type_list an associative array with machine names as keys and friendly names as values.

acquia_lift_profiles/acquia_lift_profiles.install
```diff
@@ -180,3 +180,42 @@ function acquia_lift_profiles_update_7002(&$sandbox) {
...
+ */
+function acquia_lift_profiles_update_7003() {
+  // 1) Convert existing mappings (if not already).
+  //
+  // Example
+  // * Before migration:
+  //   array (
+  //    'content_section' => '2',
+  //    'content_keywords' => '',
+  //    'persona' => 'taxonomy_context__country',
```
katbailey: This can't be right :P

acquia_lift_profiles/acquia_lift_profiles.install
```diff
@@ -180,3 +180,42 @@ function acquia_lift_profiles_update_7002(&$sandbox) {
...
+      $vocabulary = taxonomy_vocabulary_load($vocabulary_mapping);
+      $vocabulary_mappings_updated[$type] = 'taxonomy_context' . PERSONALIZE_TARGETING_ADMIN_SEPARATOR . $vocabulary->machine_name;
+    }
+  }
+  variable_set('acquia_lift_profiles_vocabulary_mappings', $vocabulary_mappings_updated);
+
+  // 2) Enable the personalize taxonomy context module.
+  $result = module_enable(array('personalize_taxonomy_context'));
+  if (empty($result)) {
+    throw new DrupalUpdateException('Module personalize_taxonomy_context could not be enabled.');
```
katbailey: This exception will not be caught - if there is any problem enabling the module you should set an error message that the user will see after updating: 
`drupal_set_message('There was a problem enabling personalize_taxonomy module. Please ensure you have Personalize module version 7.x-1.0-rc11 or higher and enable the personalize_taxonomy_context submodule, otherwise your vocabulary mappings will no longer work.', 'error')`

acquia_lift_profiles/acquia_lift_profiles.module
```diff
@@ -147,27 +147,35 @@ function acquia_lift_profiles_page_build(&$page) {
   }
   $js_path = variable_get('acquia_lift_profiles_js_path', '');
   $account_name = variable_get('acquia_lift_profiles_account_name', '');
-  $udf_mappings = variable_get('acquia_lift_profiles_udf_mappings', array());
+  $capture_identity = variable_get('acquia_lift_profiles_capture_identity', FALSE);
+  $mappings = variable_get('acquia_lift_profiles_udf_mappings', array());
+  $acquia_lift_profiles_vocabulary_mappings = variable_get('acquia_lift_profiles_vocabulary_mappings', array());
+  if (!empty($acquia_lift_profiles_vocabulary_mappings)) {
+    $mappings['vocabulary'] = $acquia_lift_profiles_vocabulary_mappings;
```
katbailey: Why are these grouped one level deeper in the structure under 'vocabulary'? I don't see how this could work in the js. Can you add a qunit test for this? Also, it no longer makes sense to call these vocabulary mappings as they don't use taxonomy vocabularies anymore. Let's call them field mappings.

acquia_lift_profiles/acquia_lift_profiles.admin.inc
ynx: [Done](https://github.com/acquia/acquia_lift/commit/da1171c8ced3ef2bf484f580db68da8ea1507acf).

acquia_lift_profiles/acquia_lift_profiles.admin.inc
ynx: [Done](https://github.com/acquia/acquia_lift/commit/da1171c8ced3ef2bf484f580db68da8ea1507acf).

acquia_lift_profiles/acquia_lift_profiles.install
ynx: Haha no it can't be. Unlike the examples 'content_section' and 'content_keywords', the example 'persona' here is definitely not an anticipated value. I guess I should note clear that as a part of defensive practice, this is there because the hook is specifically made to work even when it is somehow applied multiple times.

acquia_lift_profiles/acquia_lift_profiles.install
ynx: - The detailed error description is much better. [Done](https://github.com/acquia/acquia_lift/commit/a0b69a1237dcc058ee9aea33a4875f54f5bd333a).
- Exception would totally be caught. And I think it was the more commonly accepted practice because a failure does not increment the version number. =D

---

exception

acquia_lift_profiles/acquia_lift_profiles.module
ynx: ## fields

1) This setup totally works haha. When I started to reduce code backwards against the new "field mappings" data flow, I found that I could completely merge into udfValues' flow, so I deleted all the added code but the current flow setup.

2) I agree it makes sense to change from vocabulary mappings to field mappings. I personally feel "field mappings" is kind of too generic but I don't have better suggestions. Done [here](https://github.com/acquia/acquia_lift/commit/43a972b673d59ff57be76ced2faf0aaede278480).

3) Yes full qunit and simpletests are absolutely required and both are coming up next.

acquia_lift_profiles/acquia_lift_profiles.install
ynx: As discussed on Slack, [updated](https://github.com/acquia/acquia_lift/commit/64539a26715b7c38e8b24d649b43867dbe2d2d30) to not to mention the "out of norm" case.

ynx:
I was rebasing and noticed that some of your [changes](https://www.drupal.org/node/2476129) on develop has slightly impacted mine. Namely, the "udfValues" here. Just wanted to especially note down the rebase resolution attempt [here](https://github.com/acquia/acquia_lift/commit/6af3f5fe9b183f9a73352911a26246dcc3814559).

ynx:
I was rebasing and noticed that some of @katbailey's [changes](https://github.com/acquia/acquia_lift/commit/c5470749c154b657ddc41279be2c6ccf23657fe0) on develop has slightly impacted mine. Namely, the "udfValues". Just wanted to especially note down the rebase resolution attempt [here](https://github.com/acquia/acquia_lift/commit/34ee3f0927d14ef248c3f82620de382fbde6af7e).

Two things to note:
1) Implication is that processEvent() is now sending the Field values as well.
2) We may not like the variable name "pageFieldValue" instead "udfValue" there.

Please let me know if anything is not okay.

ynx:
Pushed to develop. Closing.

PR #155, opened 2015-04-15
Ch 2470: Don't allow targeted variations to be removed from running campaigns

eshtadc:
- Add capabilities to delete blocks from personalized blocks on wizard page
- Limit access to remove on:
  - wizard variations subform
  - personalize blocks admin page
  - personalize elements admin page
  - personalize fields node page
  - unibar
- Add error when saving an option set with an invalid removal


acquia_lift.admin.wizard.inc
```diff
@@ -1385,10 +1408,15 @@ function _acquia_lift_personalize_campaign_wizard_variations_block(&$form, &$for
   unset($element['pblock_wrapper']['blocks']['#tree']);
 
   $pblock_id = drupal_html_id('edit-' . implode('-', $parents) . '-pblock-wrapper');
+  // Remove any duplicate indicator that Drupal may have added from seeing this
+  // id generated more than once. This can happen in the AJAX callback.
+  $dup_re = "/\\-\\-[a-z0-9]+$/";
+  $pblock_id = preg_replace($dup_re, '', $pblock_id);
```
katbailey, line 1414: Fun! :P

acquia_lift.admin.wizard.inc
```diff
@@ -1409,8 +1437,50 @@ function _acquia_lift_personalize_campaign_wizard_variations_block(&$form, &$for
 
     acquia_lift_chosenify_element($block['block']['bid'], array('acquia-lift-chosen-select-half'));
 
-    // Hide the remove button at this point.
-    $block['remove']['#access'] = FALSE;
+    // Hide the remove button if the options is currently targeted in a running
+    // campaign.
+    $option_id = isset($block['option_id']['#value']) ? $block['option_id']['#value'] : '';
+    if (!empty($option_id) && !$option_delete_allowed && acquia_lift_target_option_targeted($agent_data->machine_name, $option_id)) {
```
katbailey: Minor point but can we rename `$option_delete_allowed` to `$targeted_option_delete_allowed`? Otherwise the logic looks funny as there are scenarios where `$option_delete_allowed` is false but you still allow deleting the option.

eshtadc: Sure - that makes sense.

katbailey: Or better yet - call it `$changes_enabled` like you do below :)

acquia_lift.module
```diff
@@ -2217,6 +2338,69 @@ function acquia_lift_personalize_fields_form_element_alter(&$element, $lang) {
...
+  // If an error was not found based on option values, then check the number of
+  // options for this element against the number saved in the option set.
+  if (!$shown_error) {
+    $option_set = personalize_option_set_load($element['#attributes']['data-personalize-osid']);
+    if ($number_options < count($option_set->options)) {
+      form_set_error(implode('][', $element['#parents']), t('Fields for %field_title in use by personalization cannot be removed until the personalization is paused.', array(
+        '%field_title' => $field_title,
+      )));
+    }
+  }
```
katbailey, line 2403: Can you add an explanation as to why this bit is necessary? It's not clear to me, but maybe I'm just not following the logic above properly.

eshtadc, line 2403: Oh yes - this is very weird.  I tried to give an overview in the comments at 2361.  Basically - if you have personalized images and you remove one you are doing an AJAX callback with limit validation errors on it.  As a result, when the form is rebuilt it doesn't have all the information (because the stupid limit validation errors causes the form state values to only have that button instead of the full form state).  As a result - we don't have the option_id anymore and so we can't give a nice specific error message.  What we can do in its place is to just do a quick comparison of the number of options and throw an error if it decreases.  It's not as precise or helpful to the end user, but its better than nothing.

eshtadc, line 2403: We could just skip all the fanciness and only do a comparison of the number of options and leave it simpler.  I guess it was hard to let go of the work I did above it so I left it there.... but it may be better for maintainability and simplicity.

PR #157, opened 2015-04-16
Ch 2452: Move auto-personalize to review screen

eshtadc:


acquia_lift.admin.wizard.inc
```diff
@@ -702,23 +710,22 @@ function acquia_lift_personalize_campaign_wizard_review_alter(&$form, &$form_sta
...
-      }
-    }
+  // If there is only one audience then there is no explicit targeting so
+  // check for auto-targeting options.
+  $context_element = FALSE;
+  if ($count_audiences < 2) {
+    $visitor_context = isset($agent_data->data['visitor_context']) ? $agent_data->data['visitor_context'] : array();
+    $context_element = personalize_admin_build_visitor_context_select($visitor_context, TRUE, $agent_data);
+  }
+  $has_advanced = $has_tests || $context_element !== FALSE;
```
katbailey, line 733: Not sure when it would make sense to have auto-targeting but `$has_tests` would be false...?

eshtadc, line 733: Ummmm .... yeah.... so why is it working.  Let me dig here a bit.

eshtadc, line 733: Obviously it is not.... checking tests.

eshtadc, line 733: I know what's going on here... the check on $has_tests is whether to show the test options (decision_style, etc.) and the check for $context_element is actually for the visitor_contexts.  Both of those things go into the "advanced" section so they both get utilized here.

Of course this changes after Fridays' review comments anyway now that the cache decisions got moved in there.

PR #159, opened 2015-04-20
Fixing the review submit

katbailey:


acquia_lift.admin.wizard.inc
```diff
@@ -1030,6 +1030,7 @@ function acquia_lift_personalize_campaign_wizard_submit(&$form, &$form_state) {
 
   // Run the base submit form handling.
   $agent_data = personalize_campaign_wizard_submit_base($form, $form_state);
+  $agent_data = personalize_agent_load($agent_data->machine_name);
   $agent_instance = personalize_agent_load_agent($agent_data->machine_name);
```
eshtadc, line 1034: So the personalize_campaign_wizard_submit_base is supposed to always return the correct agent.  If it's new then it creates it, otherwise it should read it from the form.  I think the problem here is that it calls _personalize_agent_from_form_values to generate the base agent class... but that's actually resetting the data property because its looking for the options value that will no longer exist due to removing optionsForm.  Clearly one of the things I missed when ripping that out.  That function gets called from a couple of places so it should be fixed or we'll start seeing other weirdness I think.

katbailey, line 1034: Hmm, but the piece that's missing is the 'targeting' element of the data property - I don't see how that would be present in the form values here at all as this is the review form... what am I missing?

eshtadc, line 1034: Exactly.... instead of just reading from the base form - since we have extra stuff now it's like the callback should instead load the agent and then update the label based on the ever-present agent_basic_info fields rather than loading directly from the agent_basic_info fields and assuming that they encompass everything.

katbailey, line 1034: You mean like this https://github.com/acquia/personalize_module/compare/CH-2470...fix-review-submit?expand=1 ?

eshtadc, line 1034: Yeah... but we should probably still think about _personalize_agent_from_form_values given that there is never going to be that "options" form there anymore.  I think that part is just to be removed.  Want me to look more into it as a follow-up?

katbailey, line 1034: Sorry, I'm being so slow - you mean just removing the bit where it sets the data property? Like the second hunk here: https://github.com/acquia/personalize_module/compare/workflow...fix-review-submit?expand=1

eshtadc, line 1034: Yup - I think it's that simple.

PR #160, opened 2015-04-20
Add icons to "Add goal".

Zarabadoo:
- Preface the text with "+".
- Add a "x" to the right of the title bar when the card is expanded.


css/acquia_lift.admin.css
```diff
@@ -300,6 +300,31 @@ tr.acquia-lift-report-control {
...
+  width: 1.4rem;
+  height: 1.4rem;
+  line-height: 1em;
+  vertical-align: text-bottom;
+}
+.acquia-lift-campaign-add .el-card .el-card__header .el-card__title[aria-controls]:before {
+  content: url('../images/actions__add.svg');
+}
+.acquia-lift-campaign-add .el-card .el-card__title:after {
+  content: url('../images/actions__close.svg');
```
eshtadc, line 320: I think any other spot that we've used svg graphics we've still kept a fallback to png - however looking at the support map I guess that isn't really necessary anymore :-)

eshtadc:
Looks ready to merge into workflow.  Thanks!

eshtadc:
Forgot - you can use https://www.drupal.org/node/2430417 in the commit message.  It's kind of a catch-all story.

css/acquia_lift.admin.css
Zarabadoo, line 320: The admin interfaces should be fine with svg only. The audience will not see these images.

eshtadc:
This has been merged into workflow along with a slew of other user-facing improvements.

PR #161, opened 2015-04-20
Shine: Changes from targeting workflow review

eshtadc:
I just re-cherry-picked this all on top of CH-2452... so I my have a merge error - will verify later... but at least you can get started taking a look.


acquia_lift.admin.wizard.inc
```diff
@@ -1936,11 +1960,11 @@ function _acquia_lift_personalize_campaign_wizard_review_tests($agent_data, $ele
     '#collapsible' => TRUE,
     '#collapsed' => TRUE,
   );
-  $decision_style = isset($agent_data->data['decision_style']) ? $agent_data->data['decision_style'] : 'adaptive';
+  $decision_style = isset($agent_data->data['decision_style']) ? $agent_data->data['decision_style'] : 'random';
```
katbailey: Oh? We're now defaulting to test-only?

PR #165, opened 2015-04-24
Fixing batch syncing

katbailey:


acquia_lift.batch.inc
```diff
@@ -6,19 +6,55 @@
...
   $operations = array();
   foreach ($agents as $agent) {
-    $operations += acquia_lift_get_sync_operations_for_agent($agent);
+    try {
+      $operations += acquia_lift_get_sync_operations_for_agent($agent);
+    }
+    catch (AcquiaLiftException $e) {
+      // Something went wrong getting all necessary batch operations for this
+      // agent, just skip it and sync the next one.
+      drupal_set_message('Could not sync the' . $agent->machine_name . ' personalization');
```
eshtadc: Technically this should probably be translatable.

acquia_lift.batch.inc
```diff
@@ -112,12 +149,53 @@ function acquia_lift_batch_finished($success, $results, $operations) {
...
+/**
+ * Returns any error that was set during batch syncing.
+ *
+ * Also unsets the session variable used to track errors. This function should
+ * only ever be called by acquia_lift_batch_sync_tests_for_agent().
+ *
+ * @return string|NULL
+ *   The most recently set error or NULL if none was set
+ */
+function _acquia_lift_batch_get_error() {
```
eshtadc, line 178: Is there a reason to only hold on to the most recent error rather than all errors?

eshtadc:
If there are errors thrown for not having goals or variation sets for this agent, then we should also check for that in the errors() function in the acquia_lift_target agent class so that it can be shown in "Let's fix this" on the review page.

tests/acquia_lift.test
```diff
@@ -2519,6 +2520,45 @@ class AcquiaLiftWebTestCampaignWizard extends AcquiaLiftWebTestBase {
     $this->assertOptionSelected('edit-audiences-everyone-else-assignment', $first_option_set->options[0]['option_id']);
   }
 
+
+  /**
+   * Test doing a straight test, i.e. all variations assigned to "Everyone Else".
+   */
+  function testNoTargeting() {
```
eshtadc, line 2527: Is this test new?  I could have sworn there was something like this already.  It looks very similar to testTargeting().

acquia_lift.batch.inc
katbailey, line 178: Actually, this is dumb as I don't even do anything with the message. Going to change it so it just sets the session var to TRUE, no message needed.

tests/acquia_lift.test
katbailey, line 2527: I actually added this when I was doing this fix: https://github.com/acquia/personalize_module/commit/22ae8c8fcafb839427da576a27767a5bc9bdcfef Figured it was good to have a test for when you step through the wizard in the most basic way possible to ensure that things happens when they're supposed to happen.

PR #171, opened 2015-04-30
CH-2587 Adding wizard bar logic to acquia_lift module.

ynx:


css/acquia_lift.admin.css
```diff
@@ -503,7 +503,7 @@ li.acquia-lift-targeting-draggable.ui-sortable-helper {
 }
 #acquia-lift-targeting-audiences .acquia-lift-draggable-variations li.acquia-lift-targeting-draggable,
 .acquia-lift-targeting-draggable {
-  background: #E6E9ED url(/sites/all/modules/contrib/acquia_lift/images/000000/drag.png) no-repeat;
+  background: #E6E9ED url(../images/000000/drag.png) no-repeat;
   background-position: left center;
```
ynx, line 507: Not directly related to current story but slotted in. ;)

Please see [CH-2710](https://backlog.acquia.com/browse/CH-2710).

acquia_lift.module
```diff
@@ -2118,21 +2118,30 @@ function acquia_lift_form_personalize_agent_option_sets_form_alter(&$form, &$for
...
+ *   1) If the personalization is already started, render the buttons;
+ *      otherwise link to the review page.
+ *   2) Legacy campaigns whose status is completed cannot be changed to any other status.
  */
 function acquia_lift_form_personalize_status_change_form_alter(&$form, &$form_state) {
   $agent_name = $form['agent_name']['#value'];
+  $agent = personalize_agent_load($agent_name);
+  $status = personalize_agent_get_status($agent_name);
+  // If the personalization is already started, render the buttons; otherwise link to the review page.
+  if (acquia_lift_is_targeting_agent($agent) && $status === PERSONALIZE_STATUS_NOT_STARTED) {
```
katbailey: Paused campaigns should also get the link to the review page instead of a start button.

eshtadc: Actually this same form is used within the process bar of the campaign details page. Should those also be redirected if they are already looking at the campaign?  @katbailey I could see this being useful from our standpoint but maybe kind of annoying for the user - but I could be persuaded.

katbailey: Hmm, if we want to make the behavior different, is there some way we can check where the form is being rendered?

ynx: @katbailey current_path(). =(

ynx: @lkhaas Do you mean the "Edit personalization" / wizard page? That "Start" button was separate. Right now in this story, this is the only user-visible change:

https://backlog.acquia.com/secure/attachment/69611/Screen%20Shot%202015-05-04%20at%2014.54.22.png

ynx:
Merged into 7.x-2.x.

PR #172, opened 2015-05-01
Ch 2607

katbailey:
This PR rips out everything to do with targeting (aka features, aka contexts) from the testing agent. It also takes a sledgehammer to the reporting code because with all the redesigns pulling it apart and smooshing it back together it had become a bit of a tangled mess.

The main reason there's not a bigger difference between the amount of red and the amount of green here is that I've left the original report class as a legacy report class. It's in its own file which will only ever be loaded when needed. There are hardly any changes to the methods in that AcquiaLiftReportLegacy class from what was in the original reporting - it is smushed into a single class (which I also did for the main reporting class as there's no longer any need for a base class and it was just confusing). I did have to add some trickery so that it could still get the targeting impact report and the context filters even though we've removed explicit support for those calls from the API. So you'll see a new makeLegacyGetRequest in the AcquiaLiftAPI class that is just for this report to be able to pull that stuff.


plugins/agent_types/AcquiaLiftAgent.inc
```diff
@@ -2151,65 +1526,88 @@ class AcquiaLiftReport extends AcquiaLiftReportBase {
...
-        if ($choice['score'] > $item['choices'][$favored_selection]['score']) {
-          $favored_selection = $i;
-        }
-      }
-      $data[$feature] = array(
-        'raw_label' => $item['label'],
-        'label' => $item['labelText'],
-        'favored_selection' =>  $item['choices'][$favored_selection]['label'],
-        'percent_traffic' => $this->formatReportPercentage($item['percentTraffic']),
-        'percent_traffic_graph' => $this->getGraphLevelClass($item['percentTraffic']),
```
eshtadc, line 2181: getGraphLevelClass can be deleted from AcquiaLiftAgent.inc I believe as well.

acquia_lift.admin.inc
```diff
@@ -287,13 +287,36 @@ function acquia_lift_admin_form_submit($form, &$form_state) {
  * @param stdClass $option_set
  *   (optional) The content variation to show in the default view.
  */
-function acquia_lift_report($form, &$form_state, $agent_data, $option_set = NULL) {
+function acquia_lift_report($form, &$form_state, $agent_data, $option_set = NULL, $nested = FALSE) {
   if (!$agent = personalize_agent_load_agent($agent_data->machine_name)) {
```
eshtadc: This seems weird and I can't remember why it's there.  Does it make sense to you?

katbailey: Yeah, I found it weird too, but then discovered that it's needed further down - it's the instantiated object and we need to call its errors() method further down. Going to rename the variable to `$agent_instance` to make it clearer and move it further down to where it's actually needed, with an explanation.

PR #173, opened 2015-05-04
CH-2663: Add a flag option to the card theme.

Zarabadoo:


theme/acquia_lift.wizard.theme.inc
```diff
@@ -16,7 +16,8 @@ function theme_acquia_lift_card($variables) {
   $footer_children = empty($element['footer']) ? array() : element_children($element['footer']);
   $card_label = isset($element['#title']) ? $element['#title'] : '';
   $header_children = isset($element['header']) ? element_children($element['header']) : array();
-  unset($element['footer']);
+  $card_flag = isset($element['#flag']) ? $element['#flag'] : '';
+unset($element['footer']);
```
eshtadc: OK... the silliest thing... indentation.

theme/acquia_lift.wizard.theme.inc
```diff
@@ -45,6 +46,9 @@ function theme_acquia_lift_card($variables) {
   }
   $html .= '</div>';
   $html .= '<div class="el-card__content">';
+  if (!empty($card_flag)) {
```
eshtadc, line 54: I'm going to have to make this flag appear dynamically in response to dragon drop.  Seems like we could either have the flag HTML structure always there and just hidden when unused by CSS or... I could pass the flag value that I intend to activate-deactivate through the theme function and then toggle it via dragon drop.  I guess one is a solution within the card and one is manipulating it from the outside.  Do you have a preference?

Zarabadoo, line 54: Be aware that this markup is in the document flow, so it will cause items to shift to accommodate the space it takes up. We could add some transitions for if it should be always present to smooth it out.

There is also an `is-active` class that could be used for dragons. Before this PR, it was being used for an expanded card and it was more intended as a "this is currently the piece to focus on" and there would only be one at a time. The cards now use `is-compact` for when the card is ... well ...  compact.

eshtadc, line 54: In the case of dragons there really are two cards active at a time, the one you are coming from and your possible target.... Specifically to the flag, more than one card could have it at a time.

Zarabadoo, line 54: It is certainly a pattern that we have not worked a lot on yet. `.is-active` Would definitely be good for the element being dragged as it has a drop shadow and will show a form of lift. In a pinch the same could be used on the targets until we think of something better.

eshtadc, line 54: For the purposes of this story... how about adding the transitions since we know we will have to deal with client-side adjustments to the visibility of the flag?  We can work out the other stuff as we go.

PR #175, opened 2015-05-08
CH-2677: As a Lift JS API consumer I want to easily get segments base…

marciocm:
…d on a tracking ID.
- Added trigger to segmentCache.store();
- Added qunit test.

### DIFF
--- a/acquia_lift_profiles/js/acquia_lift_profiles.js
+++ b/acquia_lift_profiles/js/acquia_lift_profiles.js
@@ -51,7 +51,7 @@ var _tcwq = _tcwq || [];
             }
           }
         }
-        $(document).trigger('retrievedSegments', [segments]);
+        $(document).trigger('acquiaLiftStoredSegments', [segments]);
         return visitorSegments;
       },
       'retrieve': function(settings) {
--- a/acquia_lift_profiles/js/acquia_lift_profiles.js
+++ b/acquia_lift_profiles/js/acquia_lift_profiles.js
@@ -51,6 +51,7 @@ var _tcwq = _tcwq || [];
             }
           }
         }
+        $(document).trigger('retrievedSegments', [segments]);
         return visitorSegments;
       },
       'retrieve': function(settings) {
--- a/acquia_lift_profiles/qunit/tests.js
+++ b/acquia_lift_profiles/qunit/tests.js
@@ -94,7 +94,7 @@ QUnit.module("Acquia Lift Profiles", {
   }
 });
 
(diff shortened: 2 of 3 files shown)


acquia_lift_profiles/js/acquia_lift_profiles.js
```diff
@@ -51,6 +51,7 @@ var _tcwq = _tcwq || [];
             }
           }
         }
+        $(document).trigger('retrievedSegments', [segments]);
```
katbailey: Can we rename this event to acquiaLiftStoredSegments? That namespaces it and also matches what is actually happening here, which is that we are storing segments in the cache.

acquia_lift_profiles/qunit/tests.js
```diff
@@ -177,6 +177,28 @@ QUnit.asyncTest("Get context values no cache", function( assert ) {
 
 });
 
+QUnit.asyncTest("Trigger retrievedSegments", function( assert ) {
+  expect(5);
+  var contextResult = Drupal.personalize.visitor_context.acquia_lift_profiles_context.getContext({'segment1':'segment1', 'segment2':'segment2'});
+  assert.ok(contextResult instanceof Promise);
+  var cached = Drupal.personalize.visitor_context_read('segment1', 'acquia_lift_profiles_context');
```
katbailey: This is confusing me. Maybe you could add some comments to explain what is being tested here? Should there perhaps be an assertion that cached does not contain any segments at this point?

acquia_lift_profiles/qunit/tests.js
```diff
@@ -177,6 +177,28 @@ QUnit.asyncTest("Get context values no cache", function( assert ) {
...
+  QUnit.stop();
+  Promise.all([contextResult]).then(function (loadedContexts) {
+    QUnit.start();
+    cached = Drupal.personalize.visitor_context_read('segment1', 'acquia_lift_profiles_context');
```
katbailey: Do we even need to call this here? When the promise resolves it stores the segments in the cache and triggers the event, right? So I think we just need the call to QUnit.start() in here?

acquia_lift_profiles/js/acquia_lift_profiles.js
marciocm: Renamed.

acquia_lift_profiles/qunit/tests.js
marciocm: Modified the test to only make sure acquiaLiftStoredSegments is being triggered.

PR #176, opened 2015-05-18
Ch 2190

katbailey:
Moves the logic for automatically creating a "clicks the option set" goal into acquia_lift_personalize_option_set_insert because it only applies to newly created option sets. Continues to respect the setting for fields and provides a variable that can be set to FALSE to prevent the automatic goal creation for non-fields option sets.

### DIFF
--- a/acquia_lift.module
+++ b/acquia_lift.module
@@ -2922,7 +2922,7 @@ function acquia_lift_get_campaign_details() {
 function acquia_lift_personalize_agent_list_alter(&$agents) {
   // If this is a nested agent, it should not be listed.
   foreach ($agents as $agent_name => $agent) {
-    if ($agent->plugin === 'acquia_lift') {
+    if (in_array($agent->plugin, array(ACQUIA_LIFT_TESTING_AGENT_V1, ACQUIA_LIFT_TESTING_AGENT_V2))) {
       unset($agents[$agent_name]);
     }
   }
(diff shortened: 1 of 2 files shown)


acquia_lift.module
```diff
@@ -694,47 +694,58 @@ function acquia_lift_personalize_option_set_save($option_set) {
...
-        if (visitor_actions_save_action($action)) {
-          personalize_goal_save($option_set->agent, $action_name, 1);
-        }
-      }
+  // Only create a goal if there are no goals already set up for the agent.
+  $goals = personalize_goal_load_by_conditions(array('agent' => $agent->machine_name));
+  if (empty($goals)) {
+    $js_id = personalize_stringify_osid($option_set->osid);
+    $plugin = 'link';
+    $action_label = t('Clicks @option_set', array('@option_set' => $option_set->label));
```
eshtadc, line 729: The only thing odd is that they will end up with a ton of "Clicks Variation set #1" goals which all show up in their drop-down for pre-existing goals.  Adding the personalization name may make it too long, but maybe there is something we can do to differentiate?  Not a blocker... just an observation.

daveatacquia, line 729: I would love for this to be "Clicks through 'home page banner'" where 'home page banner' is the machine name of the variation set. If there's risk of that running over, do we have some pre-existing smart truncating we can use?

eshtadc, line 729: The code here is doing that (in fact more nicely cause its the label rather than machine name).  The problem is that we are helpful and automatically suggest the name "Variation set #1", etc. so its likely that many people will have that as a variation set label in multiple campaigns.

PR #178, opened 2015-05-26
Ch 2817 - per audience test reports

katbailey:


acquia_lift.admin.inc
```diff
@@ -287,36 +304,101 @@ function acquia_lift_admin_form_submit($form, &$form_state) {
...
+    '#suffix' => '</div>',
+    '#attached' => array(
+      'css' => array(
+        drupal_get_path('module', 'acquia_lift') . '/css/acquia_lift.admin.css',
+      ),
+      'js' => array(
+        drupal_get_path('module', 'acquia_lift') . '/js/acquia_lift.admin.js',
+      ),
+      'library' => array(
+        array('acquia_lift', 'acquia_lift.reports'),
```
eshtadc, line 313: Seems like we should just add admin.css and admin.js to the reports library dependencies assuming its only used here anyway.

acquia_lift.admin.inc
```diff
@@ -1379,5 +1423,7 @@ function acquia_lift_legacy_campaign_list() {
   }
   $agents = personalize_agent_load_multiple($legacy_campaigns);
   module_load_include('inc', 'personalize', 'personalize.admin');
+  // @todo This will need to change once we have our own menu callback for listing
+  // campaigns. The regular list no longer has a report link.
```
eshtadc, line 1436: OK - one of us needs to remember to do this ;-)

acquia_lift.module
```diff
@@ -485,6 +504,16 @@ function acquia_lift_personalize_agent_type() {
 }
 
 /**
+ * Implements hook_personalize_wizard_steps_alter().
+ */
+function acquia_lift_personalize_wizard_steps_alter(&$steps) {
```
eshtadc: Can we pass the current agent data in as context and then use it to limit this reports link to when the plugin is type acquia_lift_target?  I get a link on a personalize_target campaign right now but of course it doesn't show me any reports when I click on it.

acquia_lift.admin.inc
```diff
@@ -287,36 +304,101 @@ function acquia_lift_admin_form_submit($form, &$form_state) {
...
+          '#markup'=> t('This personalization does not contain tests, no reports to show.')
+        )
+      );
+    }
+    $audience = key($audience_options);
+  }
+
+  if ($option_set = personalize_option_set_load($option_set->targeting[$audience]['osid'])) {
+    $nested_agent = personalize_agent_load($option_set->agent);
+    if (!empty($nested_agent)) {
```
eshtadc, line 360: If there is only one audience can we just change this to show the audience label instead of the dropdown:  Also - it would be nicer in the display if there were inline - pretty sure you can just add "container-inline" to the classes for the form item.

acquia_lift.admin.inc
```diff
@@ -448,7 +493,7 @@ function acquia_lift_report_custom($form, &$form_state, $agent, $agent_data, $op
   $reports = $agent->buildCampaignReports($options);
   if (!($reports['#has_data'])) {
     // This campaign hasn't been shown yet, so there is no data for reporting.
-    return array(
+    $form = array(
       'no_report' => array(
         '#markup' => t('This personalization does not yet contain information about your visitors\' website interactions. This situation generally occurs for personalizations that have either just been started or do not generate much website traffic.'),
       ),
```
eshtadc, line 508: Should return here?  Looks confusing to see this message and then see the big empty report.

acquia_lift.admin.inc
katbailey, line 1436: I'll make a note of it in the place where I make a note of such things :)

acquia_lift.module
katbailey: Ugh, totally - will fix.

acquia_lift.admin.inc
katbailey, line 360: OK, have changed it to only show the audience label if there's only one audience. Added the 'container-inline' class as well but it doesn't seem to do anything :/

acquia_lift.admin.inc
```diff
@@ -287,36 +304,110 @@ function acquia_lift_admin_form_submit($form, &$form_state) {
...
+          '#value' => $audience
+        );
+      }
+      else {
+        $form['audience_filter'] = array(
+          '#title' => t('Audience'),
+          '#type' => 'select',
+          '#options' => $audience_options,
+          '#default_value' => $audience,
+          '#attributes' => array(
```
eshtadc, line 376: Darn - this only adds to the input... it needs to be on the wrapping form element to actually have the desired effect.  Probably not worth the effort at this point for a theme function change just for this.

PR #184, opened 2015-06-15
CH-2974: Fixing batch syncing

katbailey:


acquia_lift.batch.inc
```diff
@@ -128,7 +134,47 @@ function acquia_lift_batch_process_item($item, &$context) {
...
+ * @param $operations
+ *   An array of unprocessed operations.
+ */
+function acquia_lift_batch_finish_and_schedule_campaign($success, $results, $operations) {
+  acquia_lift_batch_finished($success, $results, $operations, PERSONALIZE_STATUS_SCHEDULED);
+}
+
+/**
+ * Checks whether the batch was successful.
+ * @param $success
```
eshtadc, line 171: Almost helpful... although I don't think this throws an exception, does it?

tests/acquia_lift.test
```diff
@@ -3290,7 +3290,7 @@ class AcquiaLiftWebTestCampaignWizard extends AcquiaLiftWebTestBase {
     // Try to start the agent.
     $this->drupalPost('admin/structure/personalize/manage/' . $agent->machine_name . '/review', array(), $this->getButton('wizard_start'));
     $this->assertText('No goals have been set up for agent ' . $agent->machine_name);
-    $this->assertText('There was a problem syncing your personalization components to Lift and it cannot be started at this time.');
```
eshtadc, line 3293: Since we are just throwing the error messages directly, we should clean "agent" out of them as it is no longer on the front-end.  Could we also provide a link to the review page for the campaign/personalization so that the user can see more clearly what the problem is along with the "Let's fix this" text.

acquia_lift.batch.inc
```diff
@@ -13,16 +13,24 @@
...
+      $finish_callback = 'acquia_lift_batch_finish_and_start_campaign';
+    }
+    elseif ($status_change == PERSONALIZE_STATUS_SCHEDULED) {
+      $finish_callback = 'acquia_lift_batch_finish_and_schedule_campaign';
+    }
+    $_SESSION['acquia_lift_agent_sync'] = $agent->machine_name;
+    acquia_lift_do_batch_sync($operations, $finish_callback);
   }
   catch (AcquiaLiftException $e) {
     drupal_set_message($e->getMessage(), 'error');
```
eshtadc, line 33: Since we aren't showing anything else along with this anymore, could we prepend something to indicate that it is a Lift error that prevents synchronization?  Just seeing "No goals have been set up for agent {machine_name}" doesn't really provide enough context for the user I don't think.

PR #188, opened 2015-06-30
Ch 3124

katbailey:
OK, this one's a bit of a monster. I'll try to explain the approach I took with this refactoring. I basically wanted to make it as easy as possible to understand how reporting data gets turned into mark-up on the reports page. Currently the logic is very difficult to follow and the reporting markup expects a bunch of nested arrays of different reports.

Each report relies on certain data, so obviously we need a method responsible for getting that data from the report data source and storing it as a class property. This data can then be accessed whenever a certain piece of the report needs something from it.

When a report is created, it is always either for all goals or for a single goal. The same report never needs to care about both of these. It is simply a configuration option on the report that gets passed on to the data source.

I've abstracted out all of the markup generation functions into a base class. These functions have much simpler parameters whose structure is not dependent on the way the data were loaded from the report source. This opens the way for different reporting extensions that wish to display their data in the same way.

I've also split out the reporting classes into different files and only the interfaces are included in the .info file as the others can be loaded only when reports are being accessed.


acquia_lift.admin.inc
```diff
@@ -1081,8 +1088,9 @@ function acquia_lift_report_conversion() {
   }
 
   $reports = array();
-  if ($plugin = personalize_agent_load_agent($agent_name)) {
-    $reports = $plugin->buildConversionReport($options);
+  $plugin = personalize_agent_load_agent($agent_name);
```
eshtadc, line 1090: Weird that it also checks before this for 

```
if (!acquia_lift_is_testing_agent($agent)) {
    return array();
  }
```

acquia_lift.admin.inc
```diff
@@ -689,7 +689,7 @@ function acquia_lift_report($form, &$form_state, $agent_data) {
         );
```
eshtadc, line 691: Sorry - part of this approach of a review is that I see things that aren't in the pull request... but related.
On line 652, I'm confused by the following code:

```
if (empty($audience) && !isset($option_set->targeting[$audience]['osid'])) {
    if (empty($audience_options)) {
      return array(
        'no_report' => array(
          '#markup'=> t('This personalization does not contain tests, no reports to show.')
        )
      );
    }
    $audience = key($audience_options);
  }
```

Should the other if statement actually be connected by || rather than &&?

acquia_lift.admin.inc
```diff
@@ -711,7 +711,7 @@ function acquia_lift_report_submit($form, &$form_state) {
  * @param stdClass $option_set
  *   (optional) The content variation to show in the default view.
  */
-function acquia_lift_report_custom(&$form_state, $agent_data, $legacy = FALSE) {
+function acquia_lift_report_custom(&$form_state, $agent_data, $legacy = FALSE, $parent_agent = NULL) {
```
eshtadc, line 718: docblock fix for new parameters

includes/AcquiaLiftReportFactory.inc
```diff
@@ -0,0 +1,61 @@
...
+      'AcquiaLiftReportCache',
+      'AcquiaLiftReportDataFromFile',
+    );
+    foreach ($report_class_files as $file) {
+      module_load_include('inc', 'acquia_lift', 'includes/' . $file);
+    }
+    $agent_type = $agent_instance->getType();
+    switch ($agent_type) {
+      case ACQUIA_LIFT_TESTING_AGENT_V2:
+        $class = 'AcquiaLiftLearnReport';
```
eshtadc, line 33: These don't exist yet in this PR, right?  Just making sure I'm not missing them.

eshtadc:
Overall I think it would be helpful to have a good class introductory comment for the base report and the extending report class to make it clear what the intentions are as to the different responsibilities.  That may help us in the future when we come back to these classes to add something in and don't remember what was written in the PR intro...

acquia_lift.admin.inc
katbailey, line 1090: Hmm, yeah there's probably some tidying up to do here.

acquia_lift.admin.inc
katbailey, line 691: Totally! Fixing...

includes/AcquiaLiftReportFactory.inc
katbailey, line 33: Right, I was going to change this here to not include that stuff but I left it in for the sake of merge conflict avoidance. It won't affect anything seeing as it's not currently possible to create a V2 agent.

katbailey:
Good point - I'll add an explanation.

acquia_lift.admin.inc
katbailey, line 718: Done

acquia_lift.admin.inc
katbailey, line 1090: OK, how about this https://github.com/acquia/acquia_lift/commit/93a1450f91dd8857270185aedca670523e519a33

PR #191, opened 2015-07-07
CH-3186 Moving acquia_lift_profiles_js_path variable UI from acquia_l…

ynx:
…ift_profiles to acquia_lift module, and make it a prerequisite for creating personalizations.

### DIFF
--- a/tests/acquia_lift.test
+++ b/tests/acquia_lift.test
@@ -8,6 +8,13 @@
 class AcquiaLiftWebTestBase extends DrupalWebTestCase
 {
   /**
+   * The string to use as the owner code.
+   *
+   * @var string
+   */
+  protected $liftAOwnerCode = 'Some_valid-owner123-code';
+
+  /**
    * The string to use as the runtime API key.
    *
    * @var string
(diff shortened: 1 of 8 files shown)


tests/acquia_lift.test
```diff
@@ -473,27 +481,35 @@ class AcquiaLiftWebTestConfig extends AcquiaLiftWebTestBase {
...
   }
 
-  public function testLiftAgentAvailibility() {
-    // remove the account information
-    $lift_info = variable_get('acquia_lift_account_info', array());
+  public function testLiftAgentAvailability() {
+    // Fill all required information.
+    $this->configureAcquiaLiftAccount();
+    // Remove the account information.
     variable_del('acquia_lift_account_info');
```
katbailey, line 488: Very minor point, but why configure the account only to delete the variable straight away? :)

ynx, line 488: Haha I was trying to get the test to say "if we have the full configuration except missing this _particular_ config item, then you'll see this _particular_ error message".

PR #193, opened 2015-07-08
Ch 3076: Indicate lock-step variations display

eshtadc:
There are a number of changes to:
- Remove automatic naming of variation sets and variations
- Display all variations in static list on "What" page
- Display warning right side message on "What" page
- Update lock-step display for variations in "Who" page

Tests
- I broke the campaign tests up into sections cause they were getting so massive.  That will account for most of the changes there.
- I did add a test for the messaging on the "What" page
- There were also a lot of test changes to account for the re-organizing of the "What" form in order to have columns.


acquia_lift.admin.wizard.inc
```diff
@@ -344,6 +343,57 @@ function acquia_lift_personalize_campaign_wizard_variations_alter(&$form, &$form
...
+    '#attributes' => array(
+      'class' => array('personalize-wizard-column'),
+    ),
+  );
+
+  // If there are any errors in the multiple variation set handling for this
+  // agent then show them here while the user can do something about it.
+  if (count($option_sets) > 1) {
+    $agent_instance = personalize_agent_load_agent($agent_data->machine_name);
+    $errors = $agent_instance->getMultipleOptionSetErrors($option_sets);
```
katbailey: Ack, I don't like this. We shouldn't have to instantiate the agent to get this logic. Let's move the logic for checking option sets outside of the plugin class, the plugin's errors() method can then just call the function.

eshtadc: Well - I didn't want the plugin class to have to call outside of it's class.  What if the getMultipleOptionSetErrors is a static method that gets all it needs?

acquia_lift.admin.wizard.inc
```diff
@@ -1918,6 +1992,79 @@ function _acquia_lift_personalize_campaign_wizard_next_element($form_state, $pre
...
+  // Handle display for a single variation set.
+  if (count($option_sets) == 1) {
+    foreach($first_os->options as $delta => $option) {
+      $all_variations[$delta][0] = $labels_only ? $option['option_label'] : $option;
+    }
+    return $all_variations;
+  }
+  // Specialty handling for multivariate tests.
+  if ($handling === ACQUIA_LIFT_DECISION_MULTIVARIATE) {
+    // @todo: Produce a cartesion product of all variation options.
```
katbailey, line 2033: Pretty sure we will never do this, but it's a nice intention to have :P

eshtadc, line 2033: That's why I did a "todo" instead of doing the actual work ;-)

acquia_lift.admin.wizard.inc
katbailey: The class is already littered with external function calls so there's nothing to be gained by not adding one here - it's still not going to be unit-testable. This logic refers to a restriction that we are imposing at a higher level than just this plugin class. There's probably an argument to be made that none of this errors() stuff belongs in the class but should just be a hook, but that's for another day :P

PR #195, opened 2015-07-10
Ch 3139: Allow editing visitor actions from the campaign workflow

eshtadc:
I chose to put the logic in the unibar file because I assume that it will also be put into the logic there and that's where everything else with similar functionality resides.


acquia_lift.admin.wizard.inc
```diff
@@ -1212,6 +1260,15 @@ function acquia_lift_personalize_campaign_wizard_goals_ajax_delta($form, &$form_
 }
 
 /**
+ * Ajax handler for updating an existing goal
+ */
+function acquia_lift_personalize_campaign_wizard_goals_ajax_existing($form, &$form_state) {
+  drupal_get_messages();
```
katbailey, line 1266: Why?

eshtadc, line 1266: Any little messages like "There is a new drupal security version", etc. will get shown within the context of the returned ajax - so basically they pop up in the middle of the page.  Since they would have seen any important messages on the initial page load, I'm suppressing them in these callbacks.

katbailey, line 1266: Ah, fair enough.

PR #196, opened 2015-07-13
Ch 3075: Allow basic element variation set editing from workflow

eshtadc:
Allows editing the options labels and removing an option from the workflow to make the presentation similar to blocks.

### DIFF
--- a/acquia_lift.admin.wizard.inc
+++ b/acquia_lift.admin.wizard.inc
@@ -1721,6 +1721,8 @@ function _acquia_lift_personalize_campaign_wizard_block_single_save($block_conte
  *   The option set to be edited.
  */
 function _acquia_lift_personalize_campaign_wizard_element_single_save($values, $option_set) {
+  // Make sure to start with the latest option set information.
+  $option_set = personalize_option_set_load($option_set->osid);
   $option_set->label = $values['label'];
   foreach ($option_set->options as &$option) {
     $option['option_label'] = $values['options'][$option['option_id']]['option_label'];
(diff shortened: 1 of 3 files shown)


acquia_lift.admin.wizard.inc
```diff
@@ -1704,11 +1713,68 @@ function _acquia_lift_personalize_campaign_wizard_block_single_save($block_conte
...
+ * Helper submit function to submit the general and advanced information
+ * for an existing option set.
+ *
+ * @param array $values
+ *   The form state values specific to the option set.
+ * @param stdClass $option_set
+ *   The option set object submitted with the form.
+ */
+function _acquia_lift_personalize_campaign_wizard_variations_single_save_general($values, $option_set) {
+  $option_set = personalize_option_set_load($option_set->osid);
```
katbailey, line 1744: How come the option set needs to be loaded again here, but not in the _acquia_lift_personalize_campaign_wizard_element_single_save() function above?

katbailey, line 1744: Also, I'm not quite understanding the difference between these two functions or the reason they are both needed.

eshtadc, line 1744: So this is the general option set label/advanced settings that are in the element card for all plugin types.  The element one is only what is editable specific to the element card.  Block has its own for changes to block details.  Fields doesn't allow anything else, etc.

eshtadc, line 1744: You are right about the loading before saving.  I should either be passing it along within the object or saving or loading before each edit/save.

eshtadc, line 1744: Guess just loading it again is probably the way to go based on not breaking things down the road and the fact that it's all cached anyway.

PR #199, opened 2015-07-16
CH-3352 1) If there is no Personalizations qualified to be displayed …

ynx:
…in one Status, that Status' tab should not show. 2) If there is no Personalizations qualified to be displayed on the page, that page should not be showing the list table, but rather the initial "create personalization" message.

### DIFF
--- a/acquia_lift.admin.inc
+++ b/acquia_lift.admin.inc
@@ -257,7 +257,17 @@ function acquia_lift_agent_list($agents = array(), $legacy = FALSE) {
     $agents = personalize_agent_load_multiple(array(), array(), FALSE, TRUE, 'acquia_lift_sort_agent_started');
   }
 
-  if (empty($agents)) {
+  $sorted_agents = array();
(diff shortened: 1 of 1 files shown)


acquia_lift.admin.inc
```diff
@@ -257,7 +257,17 @@ function acquia_lift_agent_list($agents = array(), $legacy = FALSE) {
...
+  $sorted_agents = array();
+  foreach ($agents as $agent) {
+    // If this is a nested agent, it should not be listed.
+    if (!$legacy && in_array($agent->plugin, array(ACQUIA_LIFT_TESTING_AGENT_V1, ACQUIA_LIFT_TESTING_AGENT_V2))) {
+      continue;
+    }
+    $agent_status = personalize_agent_get_status($agent->machine_name);
+    $sorted_agents[$agent_status][] = $agent;
+  }
+
```
ynx, line 269: Do ALL the agent-filterings before configuring the form.

PR #203, opened 2015-07-24
Ch 3266: Do not automatically assign variations to audiences.

eshtadc:


plugins/agent_types/AcquiaLiftTarget.inc
```diff
@@ -128,7 +142,12 @@ class AcquiaLiftTarget extends PersonalizeAgentBase implements PersonalizeExplic
         '!what_url' => url('admin/structure/personalize/manage/' . $this->getMachineName() . '/variations'),
       ));
     }
-    $errors += $this->getMultipleOptionSetErrors($option_sets);
```
katbailey, line 131: What was that still doing there? :P

eshtadc, line 131: The function it calls utilizes the lock_step validation function in the module (outside of the class).  I had thought I moved it all back too, though.

PR #204, opened 2015-07-24
Ch 3265: UX changes from research study

eshtadc:
Add more instructions and styling around audience weight
Show variation set name in variations lists on what and who pages
Move advanced settings to who page
- Only shown for tests
- Removed cache decisions checkbox

### DIFF
--- a/acquia_lift.admin.wizard.inc
+++ b/acquia_lift.admin.wizard.inc
@@ -1811,7 +1811,9 @@ function acquia_lift_personalize_campaign_wizard_targeting_validate(&$form, &$fo
         continue;
       }
       $validateAudienceFields($audience['details'], 'audiences[' . $type_id . '][' . $audience_id . '][details][', $is_new, $audience['details']['mapping']['contexts']);
-      $has_tests = $has_tests || count(explode(',', $audience['assignment_order'])) > 1;
+      if (isset($audience['assignment_order'])) {
+        $has_tests = $has_tests || count(explode(',', $audience['assignment_order'])) > 1;
+      }
     }
   }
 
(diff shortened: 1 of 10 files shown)


plugins/agent_types/AcquiaLiftAgent.inc
```diff
@@ -128,8 +128,8 @@ class AcquiaLiftAgent extends PersonalizeAgentBase implements PersonalizeAgentGo
       // Acquia Lift takes the control rate as a number between 0 and 1.
       $acquia_lift_control_rate = $this->data['control_rate'] / 100;
     }
-    if (isset($this->data['explore_rate']) && isset($this->data['decision_style'])) {
-      if ($this->data['decision_style'] === 'adaptive') {
+    if (!isset($this->data['decision_style']) || $this->data['decision_style'] === 'adaptive') {
```
ynx: Does this new line resolve to as decision_style is default to "adaptive"? It seems if the decision_style is not set, then we should default to "random", which does not care about the explore rate.

eshtadc: Hmm - this line is backwards... it should be set to random.
Should read:
if (isset($this->data['decision_style']) && $this->data['decision_style'] === 'adaptive')) {
.....
}

acquia_lift.admin.wizard.inc
```diff
@@ -1090,9 +1125,7 @@ function acquia_lift_personalize_campaign_wizard_review_alter(&$form, &$form_sta
     // Add audiences to the summary section.
   $agent_data = $form['#agent'];
 
-  // If the agent has embedded tests then include the test settings form before
-  // the summary.
-  $count_audiences = 0;
+  // Add MVT and audience counts to the overview.
   $has_tests = FALSE;
```
ynx: Doesn't look like we are using "$has_tests" any more. (and line 1132)

My IDE also tells me Line 1147 $targeting = array() is redundant.

PR #214, opened 2015-10-03
CH-3395 Engagement score and global value 1) store to and retrieve fr…

ynx:
…om database, 2) send to frontend JavaScript settings, and 3) send to Lift Web.


acquia_lift_profiles/acquia_lift_profiles.admin.inc
```diff
@@ -6,6 +6,86 @@
...
+        drupal_get_path('module', 'acquia_lift') . '/css/acquia_lift.admin.css',
+      ),
+    ),
+  );
+
+  $form['lift_event_values'] = array(
+    '#tree' => TRUE,
+    '#theme' => 'acquia_lift_profiles_lift_event_value_form_table',
+  );
+  $serialized_lift_event_values = variable_get('acquia_lift_profiles_lift_event_values', 'a:0:{}');
```
ynx: There were 3 possible ways to store it and I ended trying all three. Storing in variable seemed to be the best one because:
- Storing in visitor_actions_actions table's data field blob is not okay because we want to have lift event values even for the stock / system actions such as user_login and form_submit, and for those do not have a record in the table to piggy-back on. Also it requires "hacking" the visitor_actions code to retain the data in the data-field and prevent it being overwritten when the action is been updated somewhere else.
- Storing in a new table acquia_lift_profiles_actions is an inferior approach because:
  - More code, one more table.
  - For performance reasons, we still need to pulling up the whole table every time we need to use it (see acquia_lift_profiles_visitor_actions_page_build()).

katbailey: variable_get/variable_set take care of serialization for you

ynx: Updated.

acquia_lift_profiles/acquia_lift_profiles.admin.inc
```diff
@@ -6,6 +6,84 @@
...
+}
+
+/**
+ * Submit handler for the lift event_values configuration form.
+ */
+function acquia_lift_profiles_lift_event_values_form_submit($form, &$form_state) {
+  $lift_event_values = $form_state['values']['lift_event_values'];
+  // No additional sanitization only because all the inputs are through element_validate_integer_positive()
+  variable_set('acquia_lift_profiles_lift_event_values', $lift_event_values);
+}
```
katbailey, line 70: Since there's no custom logic required here and the form values correspond exactly to the variable that should be stored, it might be possible to use system_settings_form()?

ynx, line 70: system_settings_form() indeed is the better tool here. Updated in efc51975034bf34259092cc28f770ebf004debcc.

PR #215, opened 2015-10-13
DCE-198: Perform t-tests between all pairs to find a winner, includes…

katbailey:
… tests


includes/AcquiaLiftLearnReport.inc
```diff
@@ -340,23 +309,67 @@ class AcquiaLiftLearnReport extends AcquiaLiftReportBase {
...
+      $choice_ids = array_keys($data);
+
       foreach ($data as $choice_id => &$values) {
+        // We impose a threshold of a minimum number of decisions before
+        // presenting a confidence value.
+        if ($values['count'] >= self::MIN_DECISION_THRESHOLD) {
+          // We need an array that holds all *but* this variation.
+          $other_variations = drupal_map_assoc($choice_ids);
+          unset($other_variations[$choice_id]);
+          // Kep track of how many other variations this variation beats.
```
ynx: "Keep".

includes/AcquiaLiftLearnReport.inc
```diff
@@ -340,23 +309,67 @@ class AcquiaLiftLearnReport extends AcquiaLiftReportBase {
...
+            if (empty($overall_confidence) || $confidence < $overall_confidence) {
+              $overall_confidence = $confidence;
+            }
+            $number_beaten++;
+          }
+          if ($number_beaten === count($other_variations)) {
+            $this->interpretedResults['winner'] = array_search($choice_id, $choice_ids);
+            $this->interpretedResults['confidence'] = $overall_confidence;
+          }
+        }
```
ynx, line 341: Personally, I would refactor the if statement (i.e. $values['count'] >= self::MIN_DECISION_THRESHOLD) and its content to put away the logic in a separate private function.

includes/AcquiaLiftLearnReport.inc
```diff
@@ -340,23 +309,67 @@ class AcquiaLiftLearnReport extends AcquiaLiftReportBase {
...
+        // It only makes sense to calculate lift over default if there has been
+        // at least one decision for each variation and we have a non-zero value
+        // for the mean of the default.
+        if (!$values['control'] && $control_variation['count'] && $control_variation['mean'] && $values['count']) {
+          $lift_default = (($values['mean'] - $control_variation['mean'])/$control_variation['mean']);
```
ynx: Totally nitpicking here - we don't need the outer brackets. =p

tests/acquia_lift_v2.test
```diff
@@ -850,6 +850,52 @@ debug($report['experiment']['reports']['conversion']['summary']);
...
+    $report_src = new AcquiaLiftLearnReportFromFile("/$path/tests/test_report_v2_winner_at_90.json");
+    $report = new AcquiaLiftLearnReport($agent_instance, array('confidence_measure' => 95), $report_src);
+    $this->assertNull($report->getOverallWinner());
+    $this->assertFalse($report->getOverallConfidence());
+    $report_src = new AcquiaLiftLearnReportFromFile("/$path/tests/test_report_v2_winner_at_90.json");
+    $report = new AcquiaLiftLearnReport($agent_instance, array('confidence_measure' => 90), $report_src);
+    $this->assertEqual(2, $report->getOverallWinner());
+    $this->assertTrue(abs(0.935 - $report->getOverallConfidence()) < 0.001);
+  }
+
```
ynx, line 898: Tests so awesome!

includes/AcquiaLiftLearnReport.inc
katbailey, line 341: Yeah, you're right - I don't like this, will rework.

includes/AcquiaLiftLearnReport.inc
katbailey: :)

katbailey:
@daveatacquia we removed the setting you're thinking of. This is more of an internal sanity check to make sure we don't declare 99% confidence when there have been only 15 decisions or something.

includes/AcquiaLiftLearnReport.inc
katbailey, line 341: @ynx please have a look at https://github.com/acquia/acquia_lift/commit/b2616b0c756333a8c254e69f10a72972a80a34ba - hopefully this is more to your taste :) I'm still not totally happy with it but this code will be fairly short-lived (will be replaced by the new reporting api) so I don't want to spend too much time on it.

PR #217, opened 2015-11-12
DCE-155: support the new reporting API

katbailey:


js/acquia_lift.v2.js
```diff
@@ -7,7 +7,7 @@
 
 (function ($, Drupal) {
   var instance;
-  Drupal.acquiaLiftV2API = (function () {
+  Drupal.acquiaLiftAPI = (function () {
```
eshtadc: I'm still trying to wrap my head around everything - but why is this still named ".v2.js"?  Seems like it should be replacing acquia_lift.api.js.

js/acquia-lift-js-wrapper-min.js
```diff
@@ -1 +0,0 @@
...rototype.batchSend=function(){var e,t,n;n=["-","batch"],e=this.batch.concat(),this.batchStart();if(e.length>0)return t=this.opts.batchingSkipsPreflight?{_type:"json"}:{},this.send(n,t,e,!1,function(t){var n,r,i;i=[];for(n in e)r=t!=null?t[n]:void 0,i.push(e[n].cb(r!=null?r.data:void 0));return i});return},i=s(20,function(e){return e.batchSend()}),r=function(e,t,n){var r;r={agent:e[0],type:e[1],query:t,cb:n};switch(r.type){case"decisions":e.length>2&&(r.choices=e.slice(2).join("/"));break;case"goal":e.length>2&&(r.goal=e[2])}return r},a}(),e=function(){function t(t,n,r,i){var s;s=e(t,n);if(s==null)return i(null);typeof r=="number"&&(s.timeout=r),s.onload=function(){return i(s.responseText)},s.onerror=s.ontimeout=function(){return i(null)},n!=null?(t.indexOf("_type=json")>0||s.setRequestHeader("Content-Type","application/json"),s.send(JSON.stringify(n))):s.send()}var e;return e=function(e,t){var n,r;return n=t!=null?"POST":"GET",typeof XMLHttpRequest!="undefined"&&XMLHttpRequest!==null&&(r=new XMLHttpRequest),(r!=null?r.withCredentials:void 0)!=null?r.open(n,e,!0):typeof XDomainRequest!="undefined"&&XDomainRequest!==null?(r=new XDomainRequest,r.open(n,e)):r=null,r},t}()}).call(this);
```
eshtadc, line 1: Yay - I think that means there is some outdated documentation somewhere about how to generate this.

js/acquia_lift.v2.js
katbailey: Oh - yes! Definitely an oversight. There should be no concept of "v2" anywhere anymore. Was getting rid of it piece by piece so obviously some last bits still hadn't been dealt with...

acquia_lift.install
```diff
@@ -42,13 +40,17 @@ function acquia_lift_disable() {
 function acquia_lift_uninstall() {
   $vars = array(
     'acquia_lift_account_info',
-    'acquia_lift_profiles_js_path',
+    'acquia_lift_apiv2_url',
```
eshtadc: Looks like this gets deleted in an update hook and not used anymore since?

acquia_lift.module
```diff
@@ -1255,74 +1197,35 @@ EOT;
...
-        ),
-        'dependencies' => array(
-          array('system', 'jquery'),
-        ),
-      );
-    } catch (\Exception $e) {
-      // do nothing. This fails when there are no keys present.
+  try {
+    $lift_api = AcquiaLiftAPI::getInstance($account_info);
+    $libraries['acquia_lift.page']['dependencies'][] = array('acquia_lift', 'acquia_lift.v2_agent_api');
```
eshtadc: Found another v2 label in the libraries.

acquia_lift.admin.wizard.inc
```diff
@@ -2353,7 +2354,13 @@ function acquia_lift_personalize_campaign_wizard_targeting_submit(&$form, &$form
   // Set the weight higher than any of the audience options.
   foreach($form_state['values']['audiences'] as $type_id => $type) {
     foreach ($form_state['values']['audiences'][$type_id] as $audience_id => $audience) {
-      if ($audience_id === ACQUIA_LIFT_TARGETING_EVERYONE_ELSE || !isset($audience['details']['weight']) || empty($audience['details']['name'])) {
+      // The machine name of the "everyone-else" audience may have changed, if
+      // tests for hte audience have changed, so it may have a suffix, like '-2'.
```
eshtadc, line 2358: This explanation (or something similar) might be useful at the acquia_lift_get_fallback_audience_name function since that's what most of the code uses to make this determination.  Might be worth also altering this to use that function to keep all the logic in one place (although I could see how it's clearer not to right here).

acquia_lift_profiles/acquia_lift_profiles.install
```diff
@@ -35,19 +35,25 @@ function acquia_lift_profiles_disable() {
...
-  variable_del('acquia_lift_profiles_field_mappings');
-  variable_del('acquia_lift_profiles_tracked_actions');
-  variable_del('acquia_lift_profiles_lift_event_values');
-  variable_del('acquia_lift_profiles_udf_mappings');
-  variable_del('acquia_lift_profiles_identity_param');
-  variable_del('acquia_lift_profiles_identity_type_param');
-  variable_del('acquia_lift_profiles_default_identity_type');
+  $vars = array(
+  'acquia_lift_profiles_access_key',
+    'acquia_lift_profiles_account_name',
```
eshtadc: Do you really want the account information to be removed if acquia lift profiles is uninstalled?  Those variables are now in acquia lift (although not removed in that module's hook_uninstall).  In general I find that rather confusing.  I think it would be safer to rename them in an update hook to acquia_lift_ variables to avoid confusion.

eshtadc:
I can't seem to get a report generated to see... but I only see the check for the d3/rickshaw libraries in the old _acquia_lift_get_subreport.  Since acquia_lift_report still includes the acquia_lift.reports library, the check message should be shown there as well.

eshtadc:
In acquia_lift.admin.inc, acquia_lift_report_experiment line 1429 there is an unused call to personalize_agent_load_agent.

acquia_lift_profiles/acquia_lift_profiles.install
katbailey: Whoa, nice catch! This was actually an unrelated change that Shay made way before we decided to move those variables into acquia_lift. I'm reluctant to rename the variables for now, but I will move their deletion into acquia_lift's uninstall at least.

PR #218, opened 2015-12-29
LJS-99 Capture event to build user profiles.

ynx:


config/install/acquia_lift.settings.yml
```diff
@@ -1,10 +1,21 @@
...
-secret_key: ''
-account_name: ''
-customer_site: ''
-api_url: ''
-js_path: ''
-capture_identity: null
-vocabulary_mappings: null
-udf_mappings: null
-tracked_actions: null
+credential:
```
ynx, line 1: A better name here would be "account" instead of "credential", it is no effort to change sine we have full test suites now. However, Burda's already using the module with hard coding their credentials, and I don't want to spend efforts on an update hook just yet.

src/Controller/TestController.php
```diff
@@ -1,23 +1,21 @@
...
-    $build['segments'] = array(
+    $data_api = \Drupal::service('acquia_lift.service.api.data_api');
+    $segments = $data_api->getSegments();
+    $build['segments'] = [
       '#markup' => print_r($segments, TRUE),
-    );
+    ];
     return $build;
   }
 }
```
ynx, line 29: Controller not tested because it is not a part of any acceptance criteria.

src/Service/Api/DataApi.php
```diff
@@ -278,27 +204,26 @@ public function getSegments() {
...
       $this->logger->info($success_msg);
+      return;
     }
-    else {
-      $this->logger->error($fail_msg);
-      throw new AcquiaLiftDataConnectorException($fail_msg);
-    }
+    $this->logger->error($fail_msg);
+    throw new DataApiException($fail_msg);
   }
```
ynx, line 227: In case you've noticed there is no more "else" keyword in this code base - this approach is an alternative and I don't plan to actively advocate for this guideline at Acquia.

I do have an upcoming talk at my local group on "Object Calisthenics", which one guideline is to "avoid 'else' keyword as much as possible", and I was hoping to use this code base as a part of demo to show that people almost always code better if they avoid "else".

src/Service/Api/DataApi.php
```diff
@@ -307,25 +232,24 @@ public function saveEvent($event_name, $event_type = 'OTHER') {
...
 }
```
ynx, line 253: DataApi class is not tested because it is not a part of any acceptance criteria.

src/Controller/TestController.php
```diff
@@ -1,23 +1,21 @@
 <?php
```
eshtadc, line 1: We don't need this anymore, do we?  It was just something I was playing with in development.  It could be removed from the  code and the routing config.

src/Controller/TestController.php
eshtadc, line 29: See my other comment - I think this file should be deleted as it is not in use and doesn't really provide use beyond development.

acquia_lift.install
```diff
@@ -0,0 +1,41 @@
...
+use Drupal\acquia_lift\Service\Helper\SettingsHelper;
+
+/**
+ * Implements hook_uninstall().
+ *
+ * Remove cache directory if module is uninstalled.
+ */
+function acquia_lift_uninstall() {
+  // Purging all Acquia Lift settings and js libraries.
+  drupal_flush_all_caches();
```
eshtadc: It seems like this should be unnecessary as the asset caches and config caches are cleared during the module uninstall process.  https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Extension%21ModuleInstaller.php/function/ModuleInstaller%3A%3Auninstall/8

ynx: Yes you are right. I uninstalled the module without the uninstall hook and saw full page regen. Removing the uninstall hook.

src/Controller/TestController.php
ynx, line 1: Can we keep this controller around? For me (or a Ready member) it is a convenient and reliable way to verify the credentials are correct. I think we should remove it once we have built the validation into admin settings form's submit validation.

src/Controller/TestController.php
ynx, line 29: Responded in another comment.

acquia_lift.module
```diff
@@ -0,0 +1,96 @@
...
+ * @file
+ * Drupal Module: Acquia Lift
+ *
+ * Acquia Lift lets you track customers' behavior throughout their buying
+ * journey — from anonymous visitor through to loyal customer. Acquia Lift
+ * allows you to create a unified customer profile for each individual, based
+ * on their interactions with your website. Using these profiles, Acquia Lift
+ * adaptively segments customers in real-time, letting you deliver personalized
+ * content that furthers your website visitors' engagement with your brand.
+ */
```
eshtadc, line 13: Wow - holy marketing batman :)

ynx, line 13: Love if I wrote it, but the credit goes to https://docs.acquia.com/lift. =P

ynx:
By the way, to run the PHPUnit tests (example):

```
cd ~/Projects/local/drupal8/core;
../vendor/bin/phpunit --group acquia_lift;
```

src/Service/Helper/HelpMessageHelper.php
```diff
@@ -0,0 +1,75 @@
...
+   * Get help message (by route name).
+   *
+   * @param string $route_name
+   *   Route name.
+   *
+   * @return string
+   *   The help message.
+   */
+  public function getMessage($route_name) {
+    switch ($route_name) {
```
eshtadc, line 52: Should probably include the overview help here as well.  See https://api.drupal.org/api/drupal/core%21modules%21help%21help.api.php/function/hook_help/8 for more information about the route provided to match on as well a link to the standard help template.

ynx, line 52: Yes we should do that too. Done at 94c63409a56bcbbfc3ca8a87b4e02cf0174f63c3.

Do we want to go by a template? Lift module is unique as it is just a front of a much larger SaaS system, it makes more sense to me simply just to redirect audiences to our centralized help source since those are all the values are at.

src/Service/Helper/NodeTypeThumbnailFormHelper.php
```diff
@@ -0,0 +1,190 @@
...
+      $full_key = $key_prefix . $field_key;
+
+      // Keep this field, if the field has already been processed.
+      $field_hash = spl_object_hash($field_definition);
+      if (isset($this->processedFieldHashes[$field_hash])) {
+        continue;
+      }
+      $this->processedFieldHashes[$field_hash] = TRUE;
+
+      // 1) Image type.
```
eshtadc, line 161: I have a recollection of supporting media/video in version 7.  Is there a way to do that here and/or maybe we just want to provide a hook to allow different media types to be supported?

ynx, line 161: The [media entity](https://www.drupal.org/project/media_entity) module appears to be a new and in-development module. It is not part of Drupal core. To avoid explicit adaption to this module, current approach is to iteratively walk through the tree of "entity reference" (tree) nodes, and allow all "image" leaves.

Providing a hook to allow different types to be supported is certainly an option. For Burda's case, we are lucky enough to find an immediate (and decently expandable) solution. I recommend that we provide hooks in the future use cases where we see other implementations not suiting a built-in support.

src/Form/AdminSettingsForm.php
```diff
@@ -0,0 +1,399 @@
...
+  private function buildIdentityForm() {
+    $identity_settings = $this->config('acquia_lift.settings')->get('identity');
+
+    $form = [
+      '#title' => t('Identity'),
+      '#type' => 'details',
+      '#tree' => TRUE,
+      '#group' => 'data_collection_settings',
+    ];
+//    $form['capture_identity'] = [
```
eshtadc, line 158: Why is this section commented?  It looks like the login hook is implemented.

src/Form/AdminSettingsForm.php
```diff
@@ -0,0 +1,399 @@
...
+   * Remove the protocal from an URL string.
+   *
+   * @param string $url
+   *   URL.
+   *
+   * @return string
+   *   Same URL as input except with the 'http://' and 'https://' trimmed.
+   */
+  private function removeProtocal($url) {
+    return preg_replace('~^https?://~', '', $url);
```
eshtadc, line 356: Do you need to escape the "/" characters here?

qunit/acquia_lift.tests.js
```diff
@@ -0,0 +1,69 @@
+/**
+ * @file acquia_lift.tests.js
+ */
+
+QUnit.module("Acquia Lift Data Collection tests", {
+  setup: function() {
+    drupalSettings.acquia_lift.credential.account_name = 'an_account_name';
+    drupalSettings.acquia_lift.credential.customer_site = 'a_customer_site';
+    drupalSettings.acquia_lift.pageContext.post_id = 90210;
```
eshtadc, line 9: LOL - would not have pegged you for that one.

src/Controller/TestController.php
eshtadc, line 1: If we need a "Test connection" button then we should build that rather than provide something that we don't intend to maintain.  If you keep this in here - it needs to be marked as deprecated... we don't provide anything like this now - no reason to get them used to it in 8.

qunit/acquia_lift.tests.js
ynx, line 9: Hey somebody got my reference! XD

src/Form/AdminSettingsForm.php
ynx, line 356: I didn't have to because I used "~" as the delimiter instead of the usual "/". I tested this again and it appears to work well.

src/Form/AdminSettingsForm.php
ynx, line 158: Capture identity doesn't work yet. I ran out of time then and the feature was not essential. For clarity, I just commented out the attempted hook at 98285c1499371ecc697b72b2fd4c070588af2e92. We'll revisit this feature's implementation in the future.

src/Controller/TestController.php
ynx, line 1: I would personally prefer a built-in connection check at admin settings form load, utilizing the Data Api. Marking the controller deprecated for now sound good. Done @ 288b9fe53a171da2090858252877b5c8e4864eef.

src/Tests/SettingsDataTrait.php
```diff
@@ -0,0 +1,133 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\acquia_lift\Tests\SettingsDataTrait.
+ */
+
+namespace Drupal\acquia_lift\Tests;
```
eshtadc, line 8: Seems like this is duplicated for unit tests.  Is there a way to prevent the duplication?

src/Tests/SettingsTest.php
```diff
@@ -0,0 +1,134 @@
+<?php
+
```
eshtadc, line 2: Perhaps a different test class, but I would also expect a test that the thumbnail configuration form is attached to the node types edit forms - like admin/structure/types/manage/article.  I see the unit tests that handle testing the form itself, but this would be to make sure it shows where expected.

src/Tests/SettingsDataTrait.php
ynx, line 8: The duplication was intentional because unit tests and functional tests should not share data fixtures. Some major reasonings are:
- Unit tests and functional tests have different focus, so their fixture data will naturally rest at different setups and complexity.
- Fixtures are to be expanded and updated later as the tests grow, it is a good idea to limit the impact to the same type of tests in a project of this size. In the future we may even create multiple sets of fixtures to serve type of tests.

Otherwise, to eliminate duplication is as simple as changing the require_once()'s path.

src/Tests/SettingsTest.php
ynx, line 2: Yes I agree. Ideally, I would like to have the units extensively explored, and then just a poke at functional level to ensure the chain still flows. This one here simply sneaked through my todo list.

ynx, line 2: Done @ e7921c683f005aeb4fb47404c9643c690d276899.

PR #220, opened 2016-01-12
LEX-150: Use the hmac function correctly

katbailey:


includes/AcquiaLiftAPIV2.inc
```diff
@@ -470,7 +470,7 @@ class AcquiaLiftAPIV2 extends AcquiaLiftAPIBase implements AcquiaLiftLearnReport
     $message = $this->canonicalizeRequest($verb, $host, $path, $query_args, $authorization, $timestamp, $content_type, $payload_hash);
 
     // Always use binary output
-    $signature = base64_encode(hash_hmac('sha256', $this->api_private_key, (string) $message, TRUE));
+    $signature = base64_encode(hash_hmac('sha256', (string) $message, $this->api_private_key, TRUE));
```
nickveenhof: That seems to be correct. The only doubt I have is the (string), would there ever be a case where the message is not a string but a boolean (FALSE for example)? If so, then the (string) should stay but maybe we need to add a comment above as to why this is. I think, ages ago, there was a bug where it required you to convert a boolean to a string as the hash that then got returned, without the conversion, was different.

nickveenhof: Here's an explanation as to why you need type-casting in PHP to be a string explicitly: http://stackoverflow.com/questions/13884803/why-is-hash-hmac-making-different-hashes-for-these-two-large-numbers

Same thinking applies, an integer will hash differently than an empty string and since FALSE in PHP is not a string and we can't know if the message will be empty or not, we need to convert it to a string explicitly. Not required for the private key as that always needs to be a string anyway and if that is empty, it will not be correct.

nickveenhof:
I learned that the private key we get back from the service is a base64 encoded version of the secret key. We need to base64 decode it before we can use it so that is still work we need to do before we push this fix.

nickveenhof:
@katbailey, I modified the class so that it adheres to the specification. I tested it in a single page PHP script but not with the module. I doubt there is any difference. I do need to deploy an update to Dice but this can get merged if you agree with these changes.

includes/AcquiaLiftAPIV2.inc
```diff
@@ -468,9 +470,19 @@ class AcquiaLiftAPIV2 extends AcquiaLiftAPIBase implements AcquiaLiftLearnReport
 
     // Canonicalize the request
     $message = $this->canonicalizeRequest($verb, $host, $path, $query_args, $authorization, $timestamp, $content_type, $payload_hash);
+    
+    // Decode the private key. We transfer it as a base64 encoded key.
+    $api_private_key_decoded = base64_decode($this->api_private_key);
+    // Validate if the secret key is a valid base64 encoded key.
+    if ($secret_key_decoded === FALSE) {
```
katbailey: Looks like variable name is wrong here, should be `$api_private_key_decoded`

nickveenhof: That's what you get when you are not testing the module itself. Silly me

nickveenhof:
@katbailey, variable updated. Hope it's good now.

nickveenhof:
Also adhering to the latest revision of the spec which is compatible with the Acquia Auth Proxy. I tested as much as I could separately but a Drupal Module test is still required.

PR #222, opened 2016-01-27
Adding site prefix

katbailey:
This still needs tests and also some checks regarding the length of the resulting decision set and goal names, and whether we need to clean the customer site name in any way. But it would be great to get feedback on the general approach, @ynx @nickveenhof @daveatacquia

### DIFF
--- a/acquia_lift.module
+++ b/acquia_lift.module
@@ -819,6 +819,10 @@ function acquia_lift_personalize_agent_presave($agent) {
     $lift_api = AcquiaLiftAPIClientFactory::create($account, $plugin_name);
     $agent->machine_name = $lift_api->ensureUniqueAgentName($agent->machine_name, PERSONALIZE_MACHINE_NAME_MAXLENGTH);
   }
+  // Save the Lift Web site name to the test agent as it will be used as a
+  // prefix to disambiguate goals and decision sets.
+  $site_name = variable_get('acquia_lift_profiles_site_name', 'drupal');
+  $agent->data['site_name'] = $site_name;
   // Make sure the required configurations are set.
   if (!isset($agent->data['decision_style'])) {
     $agent->data['decision_style'] = 'random';
@@ -930,7 +934,13 @@ function acquia_lift_personalize_option_set_render(&$element, $option_set) {
     }
     $child_os = personalize_option_set_load($rule['osid']);
     // Make sure the agent assets are available for the child OS.
-    $settings['agent_map'][$child_os->agent] = personalize_agent_get_map_settings($child_os->agent, $child_os, $assets);
+    $nested_agent = personalize_agent_load($child_os->agent);
+    // If the test agent has a site name property, then this is used as a prefix
+    // for decision and goal names.
+    $site_name_prefix = !empty($nested_agent->data['site_name']) ? $nested_agent->data['site_name'] . '-' : '';
+    $child_settings = personalize_agent_get_map_settings($child_os->agent, $child_os, $assets);
+    $child_settings['site_name_prefix'] = $site_name_prefix;
+    $settings['agent_map'][$child_os->agent] = $child_settings;
     $settings['option_sets'] += _personalize_convert_option_set_to_js_setting($child_os);
   }
   $settings_array = array(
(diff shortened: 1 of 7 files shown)


katbailey:
Oh, also pinging @lkhaas  - the more eyes on this the better :)

js/acquia_lift_target.js
```diff
@@ -226,6 +226,11 @@ Drupal.acquiaLiftLearn = (function() {
     initializeSession();
     settings = Drupal.settings.acquia_lift_learn;
     api = Drupal.acquiaLiftV2API.getInstance();
+    for (var agent_name in Drupal.settings.acquia_lift_target.agent_map) {
+      if (Drupal.settings.acquia_lift_target.agent_map.hasOwnProperty(agent_name)) {
+        site_name_prefixes[agent_name] = Drupal.settings.acquia_lift_target.agent_map[agent_name].site_name_prefix;
```
ynx: I realize I probably am nit-picking a temporary approach or didn't understand an impediment here - are we able to just use the variable Drupal.settings.acquia_lift_target.customer_site at JavaScript level? Then we can get rid of all the updates in file "acquia_lift.module", and prefix "SITE_NAME-" or empty string to the decision id only at file "acquia_lift_target.js".

acquia_lift.module
```diff
@@ -819,6 +819,10 @@ function acquia_lift_personalize_agent_presave($agent) {
     $lift_api = AcquiaLiftAPIClientFactory::create($account, $plugin_name);
     $agent->machine_name = $lift_api->ensureUniqueAgentName($agent->machine_name, PERSONALIZE_MACHINE_NAME_MAXLENGTH);
   }
+  // Save the Lift Web site name to the test agent as it will be used as a
+  // prefix to disambiguate goals and decision sets.
```
ynx: Other than decision sets and goals, we may also want to mention personalizations/campaigns here.

js/acquia_lift_target.js
katbailey: This is about differentiating between agents whose components use a site name prefix and those that don't. I was originally going to just store a boolean like "use_site_name_prefix" or whatever and then sure, you'd just get the site name from wherever's easiest. But I figured storing the actual site name rather than a boolean was more robust.

acquia_lift.module
katbailey: They already have unique names in Dice - see just above where it calls `$lift_api->ensureUniqueAgentName`

ynx: Right, I didn't know there was a "ensureUniqueAgentName".

js/acquia_lift_target.js
ynx: Right, I see what you are doing there now. This allows the old campaigns to keep running. Also, comparing to a boolean "use_site_name_prefix", storing the actual prefix allows the campaigns to keep running even when the customer_site value is changed in the future. Good call.

plugins/agent_types/AcquiaLiftLearn.inc
```diff
@@ -80,22 +80,25 @@ class AcquiaLiftLearn extends PersonalizeAgentBase implements AcquiaLiftLearning
   public function getAgentSyncOperations($option_sets, $new_goals, $old_goals = array()) {
     // Currently only a single option set is supported.
     $option_set = reset($option_sets);
+    // If the test agent has a site name, use this as a prefix for goals and
+    // decision sets.
+    $prefix = isset($this->data['site_name']) ? $this->data['site_name'] . '-' : '';
```
eshtadc, line 86: This is the second place I've seen the prefix defined (https://github.com/acquia/acquia_lift/blob/unique-names/acquia_lift.module#L940) - could it be encapsulated somewhere for those of us who forget if it's a hyphen or an underscore, etc.?

eshtadc:
It all looks good.  What is it that prevents the prefix from being saved with goals or variation sets that are edited?  Does this mess something up to have their name change mid-way?  I'm not wrapping my head around this part :(

plugins/agent_types/AcquiaLiftLearn.inc
katbailey, line 86: I guess I could define a SITE_NAME_PREFIX_SEPARATOR constant or something

katbailey:
Yeah, I actually thought of that late last night - pushed a fix (in the presave hook only add the property if it's a new agent)

plugins/agent_types/AcquiaLiftLearn.inc
eshtadc, line 86: It's probably not important at all... just the only thing that caught my eye and made me think it would be nice to just call acquia_lift_get_site_prefix($agent_data) (or something).  So NOT a blocker for this patch going through.

PR #224, opened 2016-04-15
Lex 484

Diodeus:
This ticket generates meta tags for the Acauia Lift credentials that will be consumed by lift.js upon load.


acquia_lift.module
```diff
@@ -65,6 +65,10 @@ function acquia_lift_page_attachments(array &$page) {
   $page_context = \Drupal::service('acquia_lift.service.context.page_context');
   $page_context->setPageContextTitle($title);
   $drupal_settings['pageContext'] = $page_context->getAll();
+  $metatags = $page_context->getMetatags();
+  $metatags_lift = $page_context->getLiftMetatags($credential_settings);
+
+
```
ynx: Two too many new lines. :D

acquia_lift.module
```diff
@@ -65,6 +65,10 @@ function acquia_lift_page_attachments(array &$page) {
   $page_context = \Drupal::service('acquia_lift.service.context.page_context');
   $page_context->setPageContextTitle($title);
   $drupal_settings['pageContext'] = $page_context->getAll();
+  $metatags = $page_context->getMetatags();
+  $metatags_lift = $page_context->getLiftMetatags($credential_settings);
```
ynx: https://github.com/acquia/acquia_lift/blob/LEX-484/src/Service/Context/PageContext.php#L80

PageContext already has $settings, we can just get it from there (handle like "field_mappings").

acquia_lift.module
```diff
@@ -74,7 +78,16 @@ function acquia_lift_page_attachments(array &$page) {
 
   // Attach JavaScript library and settings.
   $page['#attached']['library'][] = 'acquia_lift/acquia_lift';
+  $page['#attached']['library'][] = 'acquia_lift/acquia_lift_settings';
```
ynx: Why do we need this line?

acquia_lift.module
ynx: We can merge getLiftMetatags() into getMetatags(), and just call it getMetatags(). I don't think there is a need for calling (and testing) two separate functions.

src/Service/Context/PageContext.php
```diff
@@ -238,6 +246,57 @@ public function getAll() {
   }
 
   /**
+   * Get meta tags.
+   *
+   * @return array
+   *   Get meta tags.
+   */
+  public function getMetatags() {
```
ynx, line 273: We need to unit test getMetatags().

src/Service/Context/PageContext.php
```diff
@@ -238,6 +246,57 @@ public function getAll() {
...
+    $metatagAcct = [
+    '#type' => 'html_tag',
+    '#tag' => 'meta',
+      '#attributes' => [
+        'itemprop' => 'acquia_lift:account_id',
+        'content' => $credential_settings['account_name'],
+      ],
+    ];
+    $metatags[] = [$metatagAcct, 'account_id'];
+    $metatagSite = [
```
ynx: The metatags' config arrays for $metatag, $metatagAcct, and $metatagSite are identical in array structure, it looks like we are able to extract the logic into a separate private function and reduce code duplication.

src/Tests/SettingsTest.php
```diff
@@ -143,5 +143,10 @@ public function testJavaScriptAndDrupalSettings() {
     $this->assertEqual('account_name_1', $drupalSettings['acquia_lift']['credential']['account_name'], '[testJavaScriptAndDrupalSettings]: JavaScript settings contain credential.account_name.');
     $this->assertEqual(90210, $drupalSettings['acquia_lift']['pageContext']['post_id'], '[testJavaScriptAndDrupalSettings]: JavaScript settings contain pageContext.post_id.');
     $this->assertEqual('an_identity', $drupalSettings['acquia_lift']['identity']['identity'], '[testJavaScriptAndDrupalSettings]: JavaScript settings contain identity.identity.');
+
+    // Assert lift3 settings are inserted on page.
+    $this->assertRaw('acquia_lift:account_id', '[testJavaScriptAndDrupalSettings]: acquia_lift account_id is loaded on the node page.');
+    $this->assertRaw('acquia_lift:site_id', '[testJavaScriptAndDrupalSettings]: acquia_lift site_id is loaded on the node page.');
+
```
ynx: It looks like we already have the vocabularies, terms, and the config setup. Are we already able to see the field metatags on the page? We should be able to functionally test them here.

Diodeus: I'll need some help with that.

PR #226, opened 2016-04-21
LEX-504 Eliminate the extra CaptureIdentity event and capture identit…

ynx:
…y during the Content View event.

### DIFF
--- a/acquia_lift_profiles/js/acquia_lift_profiles.js
+++ b/acquia_lift_profiles/js/acquia_lift_profiles.js
@@ -116,40 +116,6 @@ var _tcwq = _tcwq || [];
   var identityCaptured = false;
 
   /**
-   * Send a captureIdentity event to ContextDB
-   *
-   * @param identifier
-   *   The identifier to be sent.
-   * @param identityType
-   *   The type of identity to pass.
-   */
-  var pushCaptureIdentity = function(identifier, identityType) {
-    if (identityCaptured) {
-      return;
-    }
-    _tcaq.push( [ 'captureIdentity', identifier, identityType ] );
-    identityCaptured = true;
-  };
-
-  /**
-   * Sends a captureIdentity event to TC using the email address from the
-   * passed in context.
-   *
-   * @param DrupalSettings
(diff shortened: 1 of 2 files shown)


acquia_lift_profiles/js/acquia_lift_profiles.js
```diff
@@ -232,12 +198,20 @@ var _tcwq = _tcwq || [];
...
           _tcwq.push(['onLoad', segmentCallback]);
-          _tcaq.push(['captureView', 'Content View', pageInfo]);
 
-          if(settings.acquia_lift_profiles.hasOwnProperty('identity')) {
-            pushCaptureIdentity(settings.acquia_lift_profiles.identity, settings.acquia_lift_profiles.identityType);
+          // Capture view.
+          var captureViewEvent = ['captureView', 'Content View', pageInfo];
+
+          // Capture identity in addition, if applicable.
+          if(!identityCaptured && settings.acquia_lift_profiles.hasOwnProperty('identity')) {
```
eshtadc, line 206: Do you also have to check settings.acquia_lift_profiles.captureIdentity like you do below on line 258.

acquia_lift_profiles/qunit/tests.js
```diff
@@ -82,12 +82,14 @@ QUnit.module("Acquia Lift Profiles", {
         }, 1000);
       }
     };
+  },
+  teardown: function() {
+    Drupal.acquia_lift_profiles.resetAll();
```
eshtadc, line 87: Yay :)

acquia_lift_profiles/js/acquia_lift_profiles.js
ynx, line 206: Sharp observation Lisa ... 

"captureIdentity" flag is set in Acquia Lift Profiles page and is the switch only for whether or not to capture user_login and user_register. The variable name is misleading as it doesn't control the query string parameters capture identity here.

eshtadc, line 206: Oooh - I get it.  OK - figured it was something like that.  👍

PR #229, opened 2016-05-31
Necessary changes to allow capture only

nickveenhof:


nickveenhof:
@ynx can I just merge this? We've tested the whole flow including the lift-capture.js and this code is making it crash in JS. Based on feedback from @acquiachrisnagy I've removed the clashing piece

ynx:
@nickveenhof Thank you for kindly bringing me in loop. Merging away is a good idea - I don't have to hold this up.

PR #231, opened 2016-06-24
Add OAuth endpoint to credentials configuration.

eshtadc:


src/Service/Context/PageContext.php
```diff
@@ -94,7 +95,9 @@ public function __construct(ConfigFactoryInterface $config_factory, EntityTypeMa
    */
   private function setPageContextCredential($credential_settings) {
     foreach (SELF::$CREDENTIAL_MAPPING as $credential_key => $tag_name) {
-      $this->pageContext[$tag_name] = $credential_settings[$credential_key];
+      if (isset($credential_settings[$credential_key]) {
```
ynx: Should be 

```
if (isset($credential_settings[$credential_key])) {
```

src/Service/Context/PageContext.php
```diff
@@ -94,7 +95,9 @@ public function __construct(ConfigFactoryInterface $config_factory, EntityTypeMa
...
+        $this->pageContext[$tag_name] = $credential_settings[$credential_key];
+      }
```
ynx, line 100: I think we should always set the pageContext, but default to empty string. This way, a meta tag (although can be empty value) is always dropped on the page.

src/Service/Context/PageContext.php
eshtadc: D'OH.  fixing...

src/Service/Context/PageContext.php
eshtadc, line 100: Well - the JS code treats anything in the meta tags as what should override admin values.  I don't want an empty oauth url to override and I don't want to force someone to enter the auth url everytime.

ynx, line 100: That's a good thought. 

On a semi-separated concern, does the JS mind if empty meta tags are simply skipped and not dropped on the page at all? If yes, should it?

ynx:
https://github.com/acquia/acquia_lift/blob/63be493c15accd834c9041c657d0a061d16778e5/src/Tests/SettingsTest.php#L101

No longer 14, should be 15 since we have added a new field. All other unit / functional tests passing. After this fix we are ready to merge. Thank you @lkhaas!

src/Service/Context/PageContext.php
eshtadc, line 100: It doesn't mind if they are optional (like this one).  It has checks in place to ensure that credentials are available, etc. (regardless of how they are added to the page).

PR #232, opened 2016-06-29
Lex 726

ynx:


acquia_lift.module
```diff
@@ -1,7 +1,6 @@
 <?php
 
 /**
- * @file
```
nickveenhof, line 4: Why removal of @file? https://www.drupal.org/coding-standards/docs#file still says it is valid.

config/schema/acquia_lift.settings.yml
```diff
@@ -12,21 +12,12 @@ acquia_lift.settings:
         customer_site:
```
nickveenhof: Can we, while we are at it, add descriptions of those settings as well?

src/Form/AdminSettingsForm.php
```diff
@@ -68,7 +61,7 @@ public static function create(ContainerInterface $container) {
   /**
    * {@inheritdoc}
    */
-  public function buildForm(array $form, FormStateInterface $form_state, Request $request = NULL) {
+  public function buildForm(array $form, FormStateInterface $form_state) {
```
nickveenhof, line 64: Please also add a whole lot more description to the settings page please. Certainly when you add a description to use default, we'd need to know what default is. I actually thought we were going to work with an advanced form and a dropdown of a region that pre-filled the variables for you depending on the region? We can start with us-east-1 as the only dropdown value available?

src/Form/AdminSettingsForm.php
```diff
@@ -346,16 +324,15 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
...
+    $settings->clear('credential.api_url');
+    if (!empty($values['api_url'])) {
+      $settings->set('credential.api_url', 'https://' . $this->removeProtocal($values['api_url']));
     }
-    $settings->set('credential.js_path', $values['js_path']);
-    $settings->set('credential.assets_url', $values['assets_url']);
     $settings->clear('credential.oauth_url');
     if (!empty($values['oauth_url'])) {
-      $settings->set('credential.oauth_url', $values['oauth_url']);
+      $settings->set('credential.oauth_url', 'https://' . $this->removeProtocal($values['oauth_url']));
```
nickveenhof: Protocal? :)

src/Service/Context/PageContext.php
```diff
@@ -59,6 +59,13 @@ class PageContext {
   ];
 
   /**
+   * JavaScript path.
+   *
+   * @var array
+   */
+  private $jsPath;
```
nickveenhof: Private? Any reason why this is private?

src/Service/Context/PageContext.php
```diff
@@ -67,7 +74,6 @@ class PageContext {
     'account_name' => 'account_id',
```
nickveenhof: Something tells me there is still a naming conflict here? Why is there account_name <> account_id and why customer_site <> site_id? If these are machine names, they should be the same.

src/Service/Context/PageContext.php
```diff
@@ -293,23 +335,37 @@ private function getMetaTagRenderArray($name, $content) {
...
-    }
-    // If still an array or empty, set title to empty.
-    if (is_array($title) || empty($title)) {
-      $this->pageContext['content_title'] = '';
-      return;
+  public function populateHtmlHead(&$htmlHead) {
+    // Attach Lift's metatags.
+    foreach ($this->pageContext as $name => $content) {
+      $renderArray = $this->getMetaTagRenderArray($name, $content);
+      $htmlHead[] = [$renderArray, $name];
```
nickveenhof, line 374: Double array? Is this needed?

src/Service/Helper/HelpMessageHelper.php
```diff
@@ -58,17 +53,8 @@ public function getMessage($route_name) {
...
-        $help_message = t('You can find more info in ') . $documentation_external_link;
-
-        // Generate Acquia Lift Web Admin link.
-        if (!empty($this->credentialSettings['api_url'])) {
-          $lift_web_link_text = t('Acquia Lift Web Admin');
-          $lift_web_link_url = Url::fromUri('https://' . $this->credentialSettings['api_url'], $link_attributes);
-          $lift_web_external_link = $this->linkGenerator->generate($lift_web_link_text, $lift_web_link_url);
-          $help_message .= t(', and control your web services settings at ') . $lift_web_external_link;
-        }
+        $help_message = t('You can find more info in ') . $documentation_external_link . t('.');
```
nickveenhof, line 56: Totally out of my league but isn't the more common thing to say "You can find more info at .."

acquia_lift.module
ynx, line 4: https://www.drupal.org/node/2665744
https://www.drupal.org/node/2304909

So @file is no longer required in D8.2 and somebody also said in one of above comments that the coding standards doc is getting updated.

The actual reasoning is that it is one-file-one-class in Symfony/D8 anyway, and the directory structure & file name should match namespace & class name already; this means there is really no advantage to have @file any more. I personally also have found myself making mistakes over defining it often. So, promptly removed.

config/schema/acquia_lift.settings.yml
ynx: Wouldn't think it is a bad idea, but I looked into others' code bases and noticed no one comments in the schema yml file. I think we just need to be more verbose in the AdminSettingsForm's form fields' description.

src/Form/AdminSettingsForm.php
ynx, line 64: I did not include descriptions in this story, because I wanted to focus the scope on the essentials and move on quickly. I agree description texts are helpful and easy, so I will add them as well.

As a customer, I do not feel that I need to know the exact default value, as long as I know I can safely leave some form fields empty. What I need to know is that I was told a list of items to be filling into the form, and I can do so without confusion or hassle. Product should also avoid tech details so customers learn less.

Pre-filled dropdown forms sounds great. It feels like a nice-to-have instead of an essential, so to avoid scope creeping I would recommend that we do this in a separate story. I can certainly create that story, wanna do that?

src/Form/AdminSettingsForm.php
ynx: Nice catch. Should be "Protocol". Will update.

src/Service/Context/PageContext.php
ynx: The reason why $jsPath is private, or in a broader sense, why in PageContext almost all class variables and functions are private, is because there is no need for any of those privates to be protected or public. i.e. OOP "minimum visibility" guideline.

src/Service/Context/PageContext.php
ynx: "account_name" and "customer_site" naming were inherited from lift2 codebase; they were also widely referred to by [Lift Web](https://github.com/acquia/acquia-lift-web/blob/cd7394c4131cd2852720325b5320fbc0e68e896f/admin/src/main/java/com/trucentric/core/domain/dice/VariationAggregation.java#L59), [legacy docs](https://docs.acquia.com/lift/drupal/configure-lift) and [support infrastructures](https://cci.acquia.com/node/4452101/edit) key / meta data. 

acquia-lift-decision-api and lift3 have changed the naming to "account_id" and "site_id".

We have already asked about this naming unification before but there was no conclusion. I'll be happy to update to match "account_id" and "site_id", but I need to make sure 1) this is where we want to go, 2) we are aware of what we are about to undertake, and 3) we have team consensus.

src/Service/Context/PageContext.php
ynx, line 374: I can try again, but judging from how I make code I am going to assume that I have already tried the "simple way" at the time of coding this.

src/Service/Helper/HelpMessageHelper.php
ynx, line 56: It's a good sanity check. The result page reads as follows:

screen shot 2016-07-05 at 16 26 52

Which I think "in" is the correct use here.

src/Service/Context/PageContext.php
ynx, line 374: Update: Code as-is is needed.

src/Form/AdminSettingsForm.php
ynx: Done at 6be36234a0b6cf69af160ac0a69d6c4041296b15.

src/Form/AdminSettingsForm.php
ynx, line 64: Description updated at 853c8e043fe8f08d4777f56ad4f1f7841364e76c.

ynx:
@eshtadc As discussed, we have stopped using js_path and started using assets_url by f807ced5826894bd3220cbe7c2ff001d44886bc0.

acquia_lift.module
nickveenhof, line 4: Ok, reasonable to me :)

config/schema/acquia_lift.settings.yml
nickveenhof: I was confused with actual schemas for config entity storage. This is good as is.

src/Form/AdminSettingsForm.php
```diff
@@ -102,50 +95,37 @@ private function buildCredentialForm() {
...
-      '#type' => 'textfield',
-      '#title' => t('JavaScript Path'),
-      '#default_value' => $credential_settings['js_path'],
       '#required' => TRUE,
     ];
     $form['assets_url'] = [
       '#type' => 'textfield',
       '#title' => t('Assets URL'),
-      '#default_value' => $credential_settings['assets_url'],
+      '#field_prefix' => 'https://',
```
nickveenhof, line 112: So we don't care what the user submits, http or https and automatically convert it to https? A little magical but neat. 👍  Might be good to also include that somewhere so that users don't get confused why their values changed?

src/Form/AdminSettingsForm.php
nickveenhof, line 64: As a customer you do want to know what it will default to, otherwise black magic defaults will always (in my experience) come and haunt you for years. Also for us, it's very important we know at any given time what the defaults are. It's fine that we have defaults in the lift.js files but we should not start to rely on those. Certainly not because those are region specific. I would like us to reconsider the dropdown for at least 2 regions, where we then move everything else into an advanced form. The dropdown is equal to the defaults with the difference that users can actually go and check what the dropdown filled in for them. I am not convinced that is a follow-up we should delay or do "later". We have the story time to do it now, so why not go ahead and do it. It's not a massive change in functionality but will, imho, massively improve the user experience since they don't have to know all these values. They only have to configure their Lift Account and Site Identifier and select a region... For every region we add, we can bring out a new release or allow anyone to fill in the values in the advanced form.

src/Form/AdminSettingsForm.php
```diff
@@ -346,30 +345,29 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
...
    *   URL.
    *
    * @return string
-   *   Same URL as input except with the 'http://' and 'https://' trimmed.
+   *   Same URL as input except with the 'http://' and 'https://' and trailing '/' trimmed.
    */
-  private function removeProtocal($url) {
-    return preg_replace('~^https?://~', '', $url);
+  private function removeProtocol($url) {
+    return rtrim(preg_replace('~^https?://~', '', $url), '/');
```
nickveenhof: Are you sure this is the best approach? I seem to remember that drupal or symfony has helper functions to aid you with such transformations? what if folks for some weird reason make a typo and make it htts:// ? :)

For example, the link module does it like this:

```
// We drop any scheme of the url.
$scheme = parse_url($url);
$search = '/' . preg_quote($scheme['scheme'] . '://', '/') . '/';
$replace = '';
$display_url = preg_replace($search, $replace, $url, 1);
```

src/Service/Context/PageContext.php
nickveenhof: If acquia-lift-decision-api and lift3 use the machine identifier site_id and account_id, we should follow those conventions in the Drupal module that utilizes both of those services. This is what was agreed on with Lift Web as well and is also used in the Lift Web REST API http://apidoc.dev.lift.acquia.com/

src/Service/Context/PageContext.php
nickveenhof, line 374: Bleh, but ok. Can we at least add some commenting above it why that is needed?

src/Service/Helper/HelpMessageHelper.php
nickveenhof, line 56: ok! 👍

src/Service/Context/PageContext.php
ynx, line 374: The render array's format is just how Drupal expects it to look like - I can comment "this is how Drupal wants it", but I don't think that's going to help the next person much. Did I miss a point or ... ?

src/Form/AdminSettingsForm.php
ynx, line 112: Yes it trims and converts, cowboy style. =P

I would consider those mechanisms already just precautions, and wouldn't expect them to catch a lot mistakes. This is because in the current process, we [generate](https://backlog.acquia.com/browse/CSO-3578?focusedCommentId=637294&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-637294) and customers receive all credential info. I wouldn't really worry about a customer going off-course from simple instructions. For this reason, I wouldn't think we need to build more.  =P

src/Form/AdminSettingsForm.php
ynx: Same as last comment - I feel 99% of the time, customers are just going to copy the exact URL from their emails and paste into the field, which is already without the protocol nor the trailing-slash. I personally feel any more fail-safe mechanisms into this front is diverting us.

At the meantime, if the customer HAS to make a mistake, the result config would look broken:

screen shot 2016-07-06 at 10 17 15

src/Form/AdminSettingsForm.php
ynx, line 64: I talked to Ted about the dropdown and reached to come consensus on how this can be implemented. In simple words, there will be a dropdown field of minimally a "Default" option, and rest are "EU Central 1" and "US East 1" (or whatever are available for selection). We are not able to eliminate the "Default" option conveniently, because the D8 module does not know about the "default" URL - it comes from the [lift.js](http://lift3assets.decision-api.acquia.com/latest/lift.js) (search for "http://api-demo-us-east-1.decision-api.acquia.com" in the file).

The story was a 3 pointer and the team barely gave it a green light. I have already expanded it to a solid 5 because I could justify the amount of tests I could add and bugs/enhancements I could bring in for the overall quality. We are adding more to this story and I now feel it is advancing into a 8, and it is starting to drag. I really think we should merge what we have already, and tackle the rest in another story (even immediately).

src/Service/Context/PageContext.php
ynx: Cool, since we already have consensus. [Done](https://github.com/acquia/acquia_lift/pull/232/commits/95f07ccabc825597cc4a99143fd0e6e3b0e7f200).

src/Form/AdminSettingsForm.php
ynx, line 64: I have created LEX-986 to capture the dropdown item, along with the rest of items that are valuable but I did not get to.

nickveenhof, line 64: :+1!

src/Form/AdminSettingsForm.php
nickveenhof, line 112: Sounds good to me, we can iterate if we encounter issues

src/Service/Context/PageContext.php
nickveenhof, line 374: Well, that's exactly the comment I would add there. But I'll leave it to you :-)

nickveenhof:
Good to be merged. Thanks for all the hard work!

src/Service/Context/PageContext.php
ynx, line 374: Yep yep just wondering if I have captured what you mean. Comment [Added](https://github.com/acquia/acquia_lift/pull/232/commits/ff8e16441fb06f77c6521514975799e8b526aebc).

src/Form/AdminSettingsForm.php
ynx, line 112: Yep yep, definitely revisit if we run into actual problems here.

ynx:
@nickveenhof Thank you for taking time reviewing in details! This provides a lot of confidence to the code that these updates are solid. *two thumbs up

PR #235, opened 2016-08-24
LEX-1104 | Support Slot Creation in Drupal UI (not merged, see PR 237)

nickveenhof:
As a Drupal Developer, I want an easy way to create a new slot in code, so that I can manage slots in lots of different ways.
The use cases here are many:
- I want to check a box on a Block config page to turn it into a Lift Slot
- I want to drag a panel pane around which automatically creates a Lift Slot
- I want to turn a Paragraph into a Lift Slot
- I want to render Slots inside a template file without creating them first

The list can go on for a very long time. These different methods need a simple way of creating and storing slots, which could look like the following:

$slot = acquia_lift_create_slot('block-id-1234');

That first function would do a few things:
- Look up in a database table local to Drupal if a Slot has already been created which maps to the string that's passed in (i.e., 'block-id-1234').
- If it exists, return all the properties of the Slot, including the unique ID of the slot needed to render in the HTML
- If it doesn't exist, create a new slot by calling the Lift APIs (Ideally using the SDK LEX-1101)
- acquia_lift_render_slot needs to render the slot id in the header to pass it with /decide request
- /decide request looks at passed in slot ids as well as slot visibility URL settings so that a drupal block can work across many pages and we don't have to worry about drupal paths vs actual URLs etc.

The second function would just render the div with the appropriate properties so it can be easily inserted into a template.

Acceptance Criteria:
- Something like the function above exist and works
- Includes automatically generated List UI from Drupal's config entities (no additional work should be done here)
- Document how to create slots in code
- Tests

Tests are missing right now. Might need some help


nickveenhof:
By accident I also had my PHPStorm to re-align using Drupal codestyle conventions. I can try to revert it but I think most changes are acceptable. Ping @ynx for review

nickveenhof:
This is also blocked from being merged until Content Hub PHP solves their http-hmac-php versioning issue. We need them to move away from http-hmac-php 3.1 so that we can pull in hmac-http-php 3.1. I discussed this with @matthewmorley and he's going to ask the team if they are ok in forking the http-hmac-php library to http-hmac-php-contenthub (as they are using v1) and move to http-hmac-php when they are ready. This unblocks this PR.

Update: No longer blocked. We use a forked version until content hub resolves the impediment on their end.

src/Tests/SettingsTest.php
```diff
@@ -29,7 +29,7 @@ protected function setUp() {
 
     $permissions = [
       'access administration pages',
-      'administer acquia lift',
+      'administer acquia_lift',
```
nickveenhof, line 32: Not sure if this is a change we can still make, but the general sense I have is that modules keep their full name if we're talking about global permissions.

ynx: Good change.

I think we can still make this change without update hook, because chances are slight that a customer site has a non-admin user that was explicitly assigned 'administer acquia lift'. At least I haven't seem any.

ynx:
Looking very nice! I am once again amazed of your speed blasting through story so quickly.

* New alignment looks great, no need to revert. You might want to let me know your setup though, because if we can both enforce the same styles.
* I can help with unit / functional tests.

Unfortunately today I was not able to get time to review this PR in details or try the functionality out, but from the looks of it, it looks great in general, and I think I'll only likely be able to pick up smaller issues. 

Will resume tomorrow.

acquia_lift.links.task.yml
```diff
@@ -0,0 +1,10 @@
+acquia_lift.admin_settings_form:
+  route_name: acquia_lift.admin_settings_form
+  title: 'Connection'
+  base_route: acquia_lift.admin_settings_form
+
+acquia_lift.slot.overview:
+  route_name: acquia_lift.slot.overview
+  title: 'Slots'
+  weight: 50
+  base_route: acquia_lift.admin_settings_form
```
ynx: Very minor - I think the convention here is to always have a new line at EOF.

acquia_lift.links.action.yml
```diff
@@ -0,0 +1,5 @@
+entity.acquia_lift_slot.add_form:
+  route_name: entity.acquia_lift_slot.add_form
```
ynx, line 2: I would try "entity.acquia_lift.slot.add_form" and "entity.acquia_lift.slot.add_form" first, which is what I have been doing with past Symfony projects. Did Drupal 8 enforced a new convention?

acquia_lift.routing.yml
```diff
@@ -4,4 +4,47 @@ acquia_lift.admin_settings_form:
     _form: '\Drupal\acquia_lift\Form\AdminSettingsForm'
     _title: 'Acquia Lift settings'
   requirements:
-    _permission: 'administer acquia lift'
+    _permission: 'administer acquia_lift'
+
+acquia_lift.slot.overview:
+  path: '/admin/config/content/acquia-lift/slots'
```
ynx, line 10: I would recommend to do "acquia_lift.slot.overview" and '/admin/config/content/acquia-lift/slot/overview'. The minimum discrepancy between the route id and actual path has helped me with greatly reducing parsings in my past.

src/Entity/Slot.php
```diff
@@ -0,0 +1,223 @@
...
+use Drupal\acquia_lift\AcquiaLiftException;
+use Drupal\acquia_lift\SlotInterface;
+use Drupal\Core\Config\Entity\ConfigEntityBase;
+use Drupal\Core\Entity\EntityStorageInterface;
+
+/**
+ * Defines the Drupal created slots.
+ *
+ * @ConfigEntityType(
+ *   id = "acquia_lift_slot",
```
ynx, line 15: I would try "acquia_lift.slot" first. 

If we agree to rename, all "acquia_lift_slot" are to be renamed to "acquia_lift.slot".

acquia_lift.routing.yml
```diff
@@ -4,4 +4,47 @@ acquia_lift.admin_settings_form:
...
+  defaults:
+    _title: 'Slots'
+    _entity_list: 'acquia_lift_slot'
+  requirements:
+    _entity_create_access: 'acquia_lift_slot'
+
+entity.acquia_lift_slot.add_form:
+  path: '/admin/config/content/acquia-lift/slots/add-facet'
```
ynx: I would use path '/admin/config/content/acquia-lift/slot/add'.

acquia_lift.routing.yml
```diff
@@ -4,4 +4,47 @@ acquia_lift.admin_settings_form:
...
+  defaults:
+    _entity_form: 'acquia_lift_slot.default'
+  requirements:
+    _entity_create_access: 'acquia_lift_slot'
+
+entity.acquia_lift_slot.edit_form:
+  path: '/admin/config/content/acquia-lift/slots/{acquia_lift_slot}'
```
ynx: I would use '/admin/config/content/acquia-lift/slot/{acquia_lift_slot}/edit'.

acquia_lift.routing.yml
```diff
@@ -4,4 +4,47 @@ acquia_lift.admin_settings_form:
...
+  defaults:
+    _entity_form: 'acquia_lift_slot.edit'
+  requirements:
+    _entity_access: 'acquia_lift_slot.edit'
+
+entity.acquia_lift_slot.enable:
+  path: '/admin/config/content/acquia-lift/slots/{acquia_lift_slot}/enable'
```
ynx, line 32: I would use '/admin/config/content/acquia-lift/slot/{acquia_lift_slot}/enable'.

acquia_lift.routing.yml
```diff
@@ -4,4 +4,47 @@ acquia_lift.admin_settings_form:
...
+  defaults:
+    _entity_form: 'acquia_lift_slot.enable'
+  requirements:
+    _entity_access: 'acquia_lift_slot.enable'
+
+entity.acquia_lift_slot.disable:
+  path: '/admin/config/content/acquia-lift/slots/{acquia_lift_slot}/disable'
```
ynx, line 39: I would use '/admin/config/content/acquia-lift/slot/{acquia_lift_slot}/disable'.

acquia_lift.routing.yml
```diff
@@ -4,4 +4,47 @@ acquia_lift.admin_settings_form:
...
+  defaults:
+    _entity_form: 'acquia_lift_slot.disable'
+  requirements:
+    _entity_access: 'acquia_lift_slot.disable'
+
+entity.acquia_lift_slot.delete_form:
+  path: '/admin/config/content/acquia-lift/slots/{acquia_lift_slot}/delete'
```
ynx, line 46: I would use '/admin/config/content/acquia-lift/slot/{acquia_lift_slot}/delete'.

acquia_lift.routing.yml
```diff
@@ -4,4 +4,47 @@ acquia_lift.admin_settings_form:
...
+  defaults:
+    _entity_form: 'acquia_lift_slot.delete'
+  requirements:
+    _entity_access: 'acquia_lift_slot.delete'
```
ynx: Newline?

acquia_lift.services.yml
```diff
@@ -11,6 +11,9 @@ services:
   acquia_lift.service.helper.help_message_helper:
     class: Drupal\acquia_lift\Service\Helper\HelpMessageHelper
     arguments: ['@config.factory', '@link_generator']
+  acquia_lift.service.helper.lift_api_helper:
+    class: Drupal\acquia_lift\Service\Helper\LiftAPIHelper
+    arguments: ['@config.factory']
```
ynx, line 16: In my past, API and RESTful helpers often spin into much larger local but contained complexity as project grows. I have made a habit of starting them off with their distinct service Directory, say "api" or "rest". I think I sometimes observe other repos do this to, so I would recommend here.

This means "acquia_lift.api.decision_api" and "Drupal\acquia_lift\api\DecisionApi".

But if you'd like to keep as-is until we actually get to refactor and split later, that's good to me too.

config/schema/acquia_lift.settings.yml
```diff
@@ -12,6 +12,12 @@ acquia_lift.settings:
         site_id:
           type: string
           label: 'Site ID'
+        public_key:
+          type: string
+          label: "Acquia Lift API Key."
```
ynx: Should be this?
"Acquia Lift API Public Key."

config/schema/acquia_lift.slot.yml
```diff
@@ -0,0 +1,38 @@
+acquia_lift.slot.*:
+  type: config_entity
+  label : 'Facet'
```
ynx: Should be "Slot"?

config/schema/acquia_lift.slot.yml
```diff
@@ -0,0 +1,38 @@
...
+      mapping:
+        pages:
+          type: sequence
+          label: 'The pages this slot applies to'
+          sequence:
+            type: string
+            description: 'The page this slot applies to'
+        condition:
+          type: string
+          label: 'show or hide the slot on gives pages.'
```
ynx: 'Show or hide the slot on given pages', and newline?

src/AcquiaLiftException.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace Drupal\acquia_lift;
+
+/**
+ * Represents an exception that occurred in some part of the Acquia Lift Module.
+ */
+class AcquiaLiftException extends \Exception {
+}
```
ynx: I would move this to Exception directory, and namespace it "Drupal\acquia_lift\Exception".

src/Entity/Controller/SlotListBuilder.php
```diff
@@ -0,0 +1,193 @@
...
+
+use Drupal\acquia_lift\AcquiaLiftException;
+use Drupal\Component\Utility\Html;
+use Drupal\Core\Config\Entity\ConfigEntityInterface;
+use Drupal\Core\Config\Entity\ConfigEntityListBuilder;
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\Core\Entity\EntityStorageInterface;
+use Drupal\Core\Entity\EntityTypeInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use GuzzleHttp\Exception\RequestException;
```
ynx: Did we use "GuzzleHttp\Exception\RequestException"?

src/Entity/Controller/SlotListBuilder.php
```diff
@@ -0,0 +1,193 @@
...
+ * Builds a listing of slot entities.
+ */
+class SlotListBuilder extends ConfigEntityListBuilder {
+
+  /**
+   * The Lift API Helper.
+   *
+   * @var \Acquia\LiftClient\Lift
+   */
+  protected $liftClient;
```
ynx, line 24: Personally, I would consider SlotListBuilder not intend to be inherited by another class, so I would make all the non-public attributes and functions, such as $liftClient, private.

src/Entity/Controller/SlotListBuilder.php
```diff
@@ -0,0 +1,193 @@
...
+   *   The entity storage class.
+   * @param  \Drupal\acquia_lift\Service\Helper\LiftAPIHelper
+   *   The Lift API Helper.
+   */
+  public function __construct(EntityTypeInterface $entity_type, EntityStorageInterface $storage) {
+    parent::__construct($entity_type, $storage);
+    try {
+      /** @var \Drupal\acquia_lift\Service\Helper\LiftAPIHelper $liftHelper */
+      $liftHelper = \Drupal::getContainer()
+        ->get('acquia_lift.service.helper.lift_api_helper');
```
ynx: I would feel very strongly against accessing container within any service. Some quick reasons include [those](http://stackoverflow.com/questions/23931321/in-symfony2-why-is-it-a-bad-idea-to-inject-the-service-container-rather-than-i), but there are really a lot more good arguments.

I would recommend to use constructor injection on "acquia_lift.service.helper.lift_api_helper". If for some reason constructor injection doesn't work in my favor (which was very rare for me), I'd try setter injection.

src/Entity/Controller/SlotListBuilder.php
```diff
@@ -0,0 +1,193 @@
...
+   *
+   * Typically, there's no need to override render(). You may wish to do so,
+   * however, if you want to add markup before or after the table.
+   *
+   * @return array
+   *   Renderable array.
+   */
+  public function render() {
+    $build['description'] = array(
+      '#markup' => $this->t("<p>This lists all the slots that were created by Drupal and synced up to the Acquia Lift Service. This does NOT list Slots created in the Acquia Lift Experience builder.</p>"),
```
ynx: Extremely minor - I don't think p tags need to be in the t().

src/Entity/Slot.php
```diff
@@ -0,0 +1,223 @@
...
+ *   label = @Translation("Slot"),
+ *   label_singular = @Translation("slot"),
+ *   label_plural = @Translation("slots"),
+ *   label_count = @PluralTranslation(
+ *     singular = "@count slot",
+ *     plural = "@count slots",
```
ynx: Should be "Slot" and "Slots"?

src/Entity/Slot.php
```diff
@@ -0,0 +1,223 @@
...
+ *     },
+ *   },
+ *   admin_permission = "administer acquia_lift",
+ *   config_prefix = "slot",
+ *   entity_keys = {
+ *     "id" = "id",
+ *     "label" = "label",
+ *     "uuid" = "uuid",
+ *     "status" = "status",
+ *   },
```
ynx, line 41: I couldn't find a reference to "entity_keys". Should "label" and "status" be in this?

src/Entity/Slot.php
```diff
@@ -0,0 +1,223 @@
...
+ *     "visibility",
+ *     "css_url"
+ *   },
+ *   links = {
+ *     "add-form" = "/admin/config/content/acquia-lift/slots/add-slot",
+ *     "edit-form" = "/admin/config/content/acquia-lift/slots/{acquia_lift_slot}",
+ *     "delete-form" = "/admin/config/content/acquia-lift/slots/{acquia_lift_slot}/delete",
+ *     "disable" = "/admin/config/content/acquia-lift/slots/{acquia_lift_slot}/disable",
+ *     "enable" = "/admin/config/content/acquia-lift/slots/{acquia_lift_slot}/enable",
+ *   }
```
ynx, line 55: Please see recommendation to path names.

src/Entity/Slot.php
```diff
@@ -0,0 +1,223 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function __construct(array $values, $entity_type) {
+    parent::__construct($values, $entity_type);
+
+    try {
+      /** @var \Drupal\acquia_lift\Service\Helper\LiftAPIHelper $liftHelper */
+      $liftHelper = \Drupal::getContainer()
+        ->get('acquia_lift.service.helper.lift_api_helper');
```
ynx: Please see above comment about container.

src/Entity/Slot.php
```diff
@@ -0,0 +1,223 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function getHtml() {
+    $css_html = "";
+    if (!empty($this->css_url)) {
+      $css_html = " data-lift-css=\"" . $this->css_url . "\"";
+    }
+    return "<div data-lift-slot=\"" . $this->id() . "\"" . $css_html . "/>";
+  }
```
ynx, line 145: Personally I'd try to keep the entity dumb and logic-less ... In a way that I'd never need to unit test a Entity, but I understand why you'd put it here.

src/Entity/Slot.php
```diff
@@ -0,0 +1,223 @@
...
+  public function getExternalSlot() {
+    $slot = new \Acquia\LiftClient\Entity\Slot();
+    $slot->setId($this->id());
+    $slot->setLabel($this->label);
+    $slot->setDescription($this->description);
+    $slot->setStatus($this->status());
+    $slot->setHtml($this->getHtml());
+    $slot->setVisibility($this->getVisibility());
+    return $slot;
+  }
```
ynx: Why not "clone" keyword? "clone" should create a brand new hard copy of the object. Two catches are: 1) original object's referenced objects are not cloned, only reference were cloned. 2) cloned object and services are detached from EntityManager and Containers.

Both catch don't apply so I think there is a good chance here we can use clone.

src/Entity/Slot.php
```diff
@@ -0,0 +1,223 @@
...
+   */
+  public function save() {
+    $return = parent::save();
+    try {
+      $this->liftClient->getSlotManager()->add($this->getExternalSlot());
+    } catch (\Exception $e) {
+      drupal_set_message(t($e->getMessage()), 'error');
+    }
+    return $return;
+  }
```
ynx, line 210: Personally I would make all Entity classes dumb, and put all logic in their EntityManager or other services, but I understand why you are implementing this way.

src/Entity/Slot.php
```diff
@@ -0,0 +1,223 @@
...
+        // deletions are already managed.
+        break;
+      }
+      try {
+        $entity->liftClient->getSlotManager()->delete($entity->id());
+      } catch (\Exception $e) {
+        drupal_set_message(t($e->getMessage()), 'error');
+      }
+    }
+  }
```
ynx, line 234: Pre-delete hook! So awesome.

src/Exception/MissingSettingsException.php
```diff
@@ -2,4 +2,5 @@
 
 namespace Drupal\acquia_lift\Exception;
 
-class MissingSettingsException extends \Exception {}
+class MissingSettingsException extends \Exception {
+}
```
ynx, line 6: Should either 1) be renamed "AcquiaLiftMissingSettingsException", or 2) Inherit from your new "AcquiaLiftException" now.

src/Form/AdminSettingsForm.php
```diff
@@ -106,6 +107,19 @@ private function buildCredentialForm() {
...
+      '#description' => 'Your Acquia Lift Public Key.',
+      '#default_value' => $credential_settings['public_key'],
+      '#required' => TRUE,
+    );
+    $form['secret_key'] = array(
+      '#type' => 'password',
+      '#description' => 'Your Acquia Lift Secret Key. Only required when setting it up for the first time or when you are changing the secret key.',
+      '#title' => t('Secret Key'),
+      '#default_value' => $credential_settings['secret_key'],
+    );
```
ynx, line 122: Before those two were deleted last time, there were more logic to protect especially the secret key. They still exist in [8.x-1.x](https://github.com/acquia/acquia_lift/blob/8.x-1.x/src/Form/AdminSettingsForm.php).

src/Form/AdminSettingsForm.php
```diff
@@ -347,6 +367,8 @@ private function setCredentialValues(Config $settings, array $values) {
     $settings->set('credential.account_id', trim($values['account_id']));
     $settings->set('credential.site_id', trim($values['site_id']));
     $settings->set('credential.assets_url', 'https://' . $this->cleanUrl($values['assets_url']));
+    $settings->set('credential.public_key', $values['public_key']);
+    $settings->set('credential.secret_key', $values['secret_key']);
```
ynx, line 371: See last comment about protecting secret key.

src/Form/SlotForm.php
```diff
@@ -0,0 +1,239 @@
...
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
+   *   The entity manager.
+   */
+  public function __construct(EntityTypeManagerInterface $entity_type_manager) {
+    $this->entityTypeManager = $entity_type_manager;
+    $this->slotStorage = $entity_type_manager->getStorage('acquia_lift_slot');
+    try {
+      /** @var \Drupal\acquia_lift\Service\Helper\LiftAPIHelper $liftHelper */
+      $liftHelper = \Drupal::getContainer()
+        ->get('acquia_lift.service.helper.lift_api_helper');
```
ynx: Please see last comment about container.

src/SlotInterface.php
```diff
@@ -0,0 +1,84 @@
+<?php
+
+namespace Drupal\acquia_lift;
```
ynx, line 3: I am guessing SlotInterface is under root because we technically could have Slot from another type of Client? 

I would recommend it to be in the same name space as "Slot".

src/SlotInterface.php
```diff
@@ -0,0 +1,84 @@
...
+
+  /**
+   * Get the Config Entity as a Acquia Lift Slot object.
+   *
+   * @return \Acquia\LiftClient\Entity\Slot
+   *   The Acquia Lift Slot object.
+   */
+  public function getExternalSlot();
+
+}
```
ynx: Newline?

src/Entity/Slot.php
```diff
@@ -0,0 +1,223 @@
...
+  public function setVisibility(Visibility $visibility) {
+    $this->visibility = $visibility->getArrayCopy();
+    return $this;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getExternalSlot() {
+    $slot = new \Acquia\LiftClient\Entity\Slot();
```
ynx: We can declare as:

use Acquia\LiftClient\Entity\Slot as LiftClientSlot;

Then do:

$liftClientSlot = new LiftClientSlot();

ynx:
Looking very nice! I am once again admiring of your speed blasting through story so quickly.
- New alignment looks great, no need to revert. You might want to let me know your setup though, because if we can both enforce the same styles.
- I can help with unit / functional tests.

Unfortunately today I was not able to get time to try the functionality out, but from the looks of it, it is great in general, and I was only able to pick up smaller issues as noted.

Will resume tomorrow.

acquia_lift.services.yml
nickveenhof: Well, I don't see the lift-sdk-php to be only Decision API, so I rather call it api/LiftApi perhaps

src/Entity/Controller/SlotListBuilder.php
nickveenhof: You don't have a choice here ;-) This is how Drupal supports that Config Entity List in the UI. I can make the liftClient private, that's not a big deal

src/Entity/Controller/SlotListBuilder.php
nickveenhof: That's the problem, the helper requires the passwords to be set, and Drupal goes crazy if it cannot instantiate the service.  If we want to inject it, it means that the connection won't be ready until you call a function to "loadCredentials" or something similar before you can start and use the service. To me, loading the service and having it throw errors if not all the credentials are there is a better DX for custom use compared to enforcing another function to happen first before we can use it. Thinking this through, I could make every function call the loadCredentials phase before use so this becomes a possibility. I'll see what I can do :)

src/Entity/Controller/SlotListBuilder.php
nickveenhof: Meh, it was like this in the example. But sure :)

acquia_lift.links.action.yml
nickveenhof: Well, the examples always follow the following convention: entity.{entity_type_id}.{action}. In our case the entity_id is acquia_lift_slot. We could rename it to just "slot" and pray no other module will ever use this. Search API uses search_api_index, Facets module uses facets_facet. I'm still in favor of keeping acquia_lift_slot and keep it as is.

acquia_lift.routing.yml
nickveenhof: same here, the entity_type_id = acquia_lift_slot so we need to keep consistency here. Also, If we make this overview, /admin/config/content/acquia-lift/slots won't go anywhere unless we allow both paths to route to the same thing. Also fine by me, but not having /admin/config/content/acquia-lift/slots and having /admin/config/content/acquia-lift/slots/overview is also not acceptable. Thoughts?

src/Entity/Slot.php
nickveenhof: As said above, I don't agree ;-)

acquia_lift.routing.yml
nickveenhof: /admin/config/content/acquia-lift/slots/add then, same reasoning. /admin/config/content/acquia-lift/slot does not exist, so I don't feel comfortable using a route where the parent does not correspond to anything. This is not a REST API :)

acquia_lift.routing.yml
nickveenhof: Wanted to fix this, but drupal actually requires this to be edit_form, convention.

acquia_lift.routing.yml
nickveenhof: Not in favor, same reasoning as above that /admin/config/content/acquia-lift/slot does not exist. Also not in line with examples and other contrib modules

src/AcquiaLiftException.php
nickveenhof: Fixed

src/Entity/Controller/SlotListBuilder.php
nickveenhof: Removed

src/Entity/Controller/SlotListBuilder.php
nickveenhof: Made this as best as I could. Please re-review. I spoke to borisson and dawehner in IRC (drupal gurus) and they suggested the solution I chose in latest commit.

src/Entity/Slot.php
nickveenhof: entity keys - mapping of the main entity keys to the entity properties we defined. For instance, when we call the label() method on the entity object, it will call the label property. I don't think any key should be in here if it ain't different.

src/Entity/Slot.php
nickveenhof: No can do, drupal convention. I tried

src/Entity/Slot.php
nickveenhof: Fixed in latest commit

src/Entity/Slot.php
nickveenhof: Semi-fixed in latest commit. The alternative is a hook, not much better.

src/Exception/MissingSettingsException.php
nickveenhof: Changed it around a bit, no longer applies

src/Form/AdminSettingsForm.php
nickveenhof: Can you take this? I don't quite follow what you said.

src/Form/SlotForm.php
nickveenhof: Fixed

src/SlotInterface.php
nickveenhof: Meh, I would leave it. Any more arguments other than feelings? ;-)

src/SlotInterface.php
nickveenhof: Fixed

src/Entity/Slot.php
nickveenhof: Fixed

src/Entity/Slot.php
nickveenhof: Tried, failed - they need to be there

acquia_lift.links.task.yml
nickveenhof: Fixed

acquia_lift.routing.yml
nickveenhof: Not in favor. Same reasoning as above

acquia_lift.routing.yml
nickveenhof: Not in favor. Same reasoning as above

acquia_lift.routing.yml
nickveenhof: Fixed

nickveenhof:
Tried to address all your comments. Thanks for the great code review!

src/SlotInterface.php
ynx: Nah, not feelings at all this time haha. It's a commonly observed convention that [an entity](https://github.com/drupal/drupal/blob/4d4be42a32d8541df122b7658330a500525c7b51/core/lib/Drupal/Core/Entity/Entity.php) and its [entity interface](https://github.com/drupal/drupal/blob/4d4be42a32d8541df122b7658330a500525c7b51/core/lib/Drupal/Core/Entity/EntityInterface.php) share the same physical directory and namespace.

The feelings part I didn't say was then we can have a completely file-free src directory. =D

src/Form/AdminSettingsForm.php
ynx: I can take this.

src/Entity/Slot.php
ynx: I was not able to find Drupal documentation yesterday, but I can try to mount arguments:
- The closest documentation I can find is [config#format](https://www.drupal.org/coding-standards/config#format). It attempts to dot-split all the layers.
- On the other hand, if we have to go for "acquia_lift_slot" id, then the current "config/schema/acquia_lift.slot.yml" would be wrong, it should be "config/schema/acquia_lift_slot.yml"; its line 1 needs to be "acquia_lift_slot.*". Also, "acquia_lift.settings" would then need to be renamed to "acquia_lift_settings" to align with "acquia_lift_slot". This in general would be against the commonly observed schema.yml patterns.

acquia_lift.links.action.yml
ynx: Moving this discussion to another review thread which is focusing on "acquia_lift_slot" id naming.

acquia_lift.routing.yml
ynx: It seems Drupal 8 core has the naming all over the place, for example [block](https://github.com/drupal/drupal/blob/6b95f9a8503b5b3a45e8e45b0805ee31f879fd9b/core/modules/block/block.routing.yml) and [block_content](https://github.com/drupal/drupal/blob/71142958f963e7cd6c3132e6337e8406fa2a17e3/core/modules/block_content/block_content.routing.yml).

So I guess it is just a matter of personal preference here. I have a feeling the setup you'll buy-in would be [dblog](https://github.com/drupal/drupal/blob/edd7889694a4f63395f12f6c5fd2d178f8017ca9/core/modules/dblog/dblog.routing.yml)'s. My only recommendation would be that we should use "slot" all across (and no "slots"), as the core conversions suggest.

Otherwise, my personally (and definitely biased) preference would go for "acquia_lift.slot.overview" and "/admin/config/content/acquia-lift/slot/overview", and avoid defining or using "/admin/config/content/acquia-lift/slot" completely, which I hope is possible. But again, personal preference is for your consideration but not an ask.

acquia_lift.routing.yml
ynx: Not sure what you are saying - my recommendation here is only to change "slots" to "slot", as per core conventions.

acquia_lift.routing.yml
ynx: Please see last comment about "slots" vs "slot".

acquia_lift.routing.yml
ynx: Please see last comment about "slots" vs "slot".

src/Entity/Slot.php
ynx: Thanks for explaining and thanks for trying to cut.

ynx:
You are very welcome @nickveenhof. Thank you for the patience polishing the code which was already fantastic, and answering questions.

nickveenhof:
Merging it in, we no longer have a library code conflict

nickveenhof:
I still need help for some of the unit and integration tests. @ynx if you have time this week?

ynx:
I have made some good progress on my current LEX-238, the UI is complete and functionality is working, but it feels like that it is going to take me at least the rest of the week to fully complete this story (see the broken down subtasks in LEX-238). Right now I don't think I will be having any bandwidth to take on more.

nickveenhof:
Reverted it all. Making a new PR and adding unaddresses comments here

src/Tests/SettingsTest.php
nickveenhof: Marking as fixed then

acquia_lift.routing.yml
nickveenhof: Marking as resolved. It follows the conventions of other contrib modules, as well as the one in the examples module. Also, if there are more that have similar opinions I'd love to hear it. I don't think this is a showstopper for now. Feel free to post a PR if you think this is super critical :)

src/Entity/Slot.php
nickveenhof: Drupal auto-detects this namespace so changing acquia_lift.slot.yml would cause for Drupal to no longer find this file. I think it takes the Classname and the module name. This makes sense as it then will be unique to that module, and the id will be unique towards the complete drupal scope. This should not be changed. Renaming is great but this will break the module. The only alternative we can do is to rename the class from Slot.php to AcquiaLiftSlot.php and then the config file becomes acquia_lift.acquia_lift_slot I think. Happy to accept pull requests if you manage to clean this up further. Marking as "not a bug" :)

acquia_lift.routing.yml
nickveenhof: Resolved this in branch LEX-1104-2. Renamed it all to /slot instead of /slots

acquia_lift.routing.yml
nickveenhof: Resolved

acquia_lift.routing.yml
nickveenhof: Resolved

nickveenhof:
Addressed all code comments as much as possible here. I'm closing this PR out in favor of https://github.com/acquia/acquia_lift/pull/237. Let's start fresh when adding the tests.

src/Form/AdminSettingsForm.php
nickveenhof: Looked at 8.x-1.x and copied the few lines that were added to the secret key and public key. There was a conditional description but I kept it static. I could not see the value add of making that conditional

acquia_lift.routing.yml
ynx: Agree to leave as-is.

src/Entity/Slot.php
ynx: Thank you for trying, Nick. Agree to leave as-is.

acquia_lift.routing.yml
ynx: "/admin/config/content/acquia-lift/slots/add" is the best.

acquia_lift.routing.yml
ynx: Got it. Thank you for trying, Nick.

acquia_lift.routing.yml
ynx: Nice nice!

acquia_lift.routing.yml
ynx: Awesome!

acquia_lift.routing.yml
ynx: Awesome!

acquia_lift.services.yml
ynx: Sounds good. *thumbs

src/Entity/Slot.php
ynx: Thank you for trying that, Nick.

src/Form/AdminSettingsForm.php
ynx: The code you add was great. Please consider this thread resolved.

src/Entity/Controller/SlotListBuilder.php
ynx: Hey thank you for attempting on this issue. Yeah resolving dependency to container has always been one of my bigger challenges in the past too. I think the solution is great for now, and I wouldn't bother with requesting anything more. Please consider this thread resolved.

PR #237, opened 2016-09-06
LEX-1104 | Create Slot UI - Second attempt

nickveenhof:
See https://github.com/acquia/acquia_lift/pull/235. Closed that one too early
As a Drupal Developer, I want an easy way to create a new slot in code, so that I can manage slots in lots of different ways.
The use cases here are many:
- I want to check a box on a Block config page to turn it into a Lift Slot
- I want to drag a panel pane around which automatically creates a Lift Slot
- I want to turn a Paragraph into a Lift Slot
- I want to render Slots inside a template file without creating them first

The list can go on for a very long time. These different methods need a simple way of creating and storing slots, which could look like the following:

$slot = acquia_lift_create_slot('block-id-1234');

That first function would do a few things:
- Look up in a database table local to Drupal if a Slot has already been created which maps to the string that's passed in (i.e., 'block-id-1234').
- If it exists, return all the properties of the Slot, including the unique ID of the slot needed to render in the HTML
- If it doesn't exist, create a new slot by calling the Lift APIs (Ideally using the SDK LEX-1101)
- acquia_lift_render_slot needs to render the slot id in the header to pass it with /decide request
- /decide request looks at passed in slot ids as well as slot visibility URL settings so that a drupal block can work across many pages and we don't have to worry about drupal paths vs actual URLs etc.

The second function would just render the div with the appropriate properties so it can be easily inserted into a template.

Acceptance Criteria:
- Something like the function above exist and works
- Includes automatically generated List UI from Drupal's config entities (no additional work should be done here)
- Document how to create slots in code
- Tests

Tests are missing right now. Might need some help


nickveenhof:
This time all existing tests are passing as well. My plan is to add the following tests:

Webtests:
Adding, Deleting, Disabling/Enabling and Listing of slots

Unit tests:
APILoader class
Slot Entity class

nickveenhof:
I victoried too early. This time the tests are actually passing in CLI mode as well. There was quite some work to figure out how to make them pass decently in CLI mode. Now I can move on to create webtests for the slots interface

PR #238, opened 2016-09-07
LEX-1122: Enable content replacement mode settings

eshtadc:


config/install/acquia_lift.settings.yml
```diff
@@ -17,4 +17,5 @@ thumbnail: []
 visibility:
   # Track, only when the page URL falls OUTSIDE OF these path patterns.
   path_patterns: "/admin\n/admin/*\n/batch\n/node/add*\n/node/*/*\n/user/*/*\n/block/*"
+content_replacement_mode: 'untrusted'
```
ynx: 1. I think a more accurate name would be "default_content_deployment_mode". 
   - I consider "deployment" over "replacement" is far more anticipated and less confusing for a new user, thus it the ideal word. I do realize changing "replacement" to "deployment" is a lot of work, so I would urge to consider the long term benefits but am okay if you don't want to change. 
   - I do I think "default" needs to be there; without "default" the literal meaning is very different than it is.
2. Currently, there is a consistency between the actual settings object and their menu representation. To keep this consistency, there should be an "advanced" parent setting, and then its "default_content_deployment_mode" child setting. This means I am recommending:
   
   advanced:
     default_content_deployment_mode: 'untrusted'

P.S. I am concerned about the "Advanced" wording, but will note in a separate comment.

config/schema/acquia_lift.settings.yml
```diff
@@ -70,6 +70,9 @@ acquia_lift.settings:
         path_patterns:
           type: string
           label: 'Path patterns'
+    content_replacement_mode:
+      type: string
+      label: 'Content replacement mode'
```
ynx: See last comment. Recommending:

```
 advanced:
   type: config_object
   label: 'Advanced'
   mapping:
     default_content_deployment_mode:
       type: string
       label: 'Default content deployment mode'
```

src/Form/AdminSettingsForm.php
```diff
@@ -317,6 +318,31 @@ private function buildThumbnailUrlForm() {
...
+    ];
+    $form['content_replacement_mode'] = [
+      '#type' => 'radios',
+      '#title' => t('Content replacement mode'),
+      '#default_value' => $settings->get('content_replacement_mode'),
+      '#options' => ['trusted' => t('Trusted'), 'untrusted' => t('Untrusted')],
+    ];
+
+    return $form;
+  }
```
ynx, line 343: I would personally recommend the following changes to "Advanced" tab:
1. In admins menu, there is already a "Data collection settings", which encapsulates all the data collection settings. We are actually introducing its sibling, "Content deployment settings", for the first time.
2. Rename "Advanced" to "Content", and list as the first tab under "Content deployment settings".
3. "Content" tab should have only one radio buttons set, which is "Default content deployment mode" of trusted / untrusted.
4. Add descriptions on trusted / untrusted mode; make aware this is a _default_ and _override-able_ value.

The reasons are:
- My old plan was to tree-group all the "data collection" options in one tab list, and the "content deployment" in another tab list.  It would be nice if I get to keep this plan, because at least two nice things come out of it:
  - Clear division and easy to find.
  - In the distance future when this pages grows too large, we split it in to 3 pages "Credential", "Data Collection", and "Content deployment".
- Choosing the default trusted mode doesn't seem to be "advanced" to me. This is one of the most requested feature as per late feedbacks, and I would consider it essential, and would expect it to be one of the first items that an admin would happily change.
- It is the best place to put down description on trusted / untrusted mode because an Admin sees this page before using the Experience Builder, and this setting has a serious impact on their personalizations' implementation on Builder.

src/Service/Context/PageContext.php
```diff
@@ -182,6 +184,19 @@ private function setPageContextCredential($credential_settings) {
...
+   *
+   * @param $settings \Drupal\Core\Config\ImmutableConfig
+   *   The lift settings values.
+   */
+  private function setPageContextConfig($settings) {
+    $replacement_mode_setting = $settings->get('content_replacement_mode');
+    if (SettingsHelper::isValidContentReplacementMode($replacement_mode_setting)) {
+      $this->pageContext['contentReplacementMode'] = $replacement_mode_setting;
+    }
+  }
```
ynx, line 197: If we make the changes as recommended above, here we would need:

`$content_settings = $settings->get('content');`
and
`$this->setPageContextContent($content_settings);`

tests/src/Unit/Traits/SettingsDataTrait.php
```diff
@@ -78,4 +78,14 @@ private function getValidVisibilitySettings() {
...
+  /**
+   * Get a valid content replacement mode
+   *
+   * @return string
+   *   A valid mode for content replacement.
+   */
+  private function getValidContentReplacementMode() {
+    return 'trusted';
+  }
 }
```
ynx, line 93: The way you unit and functional tested the new feature makes me want to bear-hug you. So good!

config/install/acquia_lift.settings.yml
eshtadc: It's not "deployment" of your content.  Particularly when we get into the realm of content hub, etc. this could lead to a lot of confusion.  It is how your content will be "replaced" during personalization.  I can see the case for nesting and I'll go ahead and add. In terms of the word "default", we could get into all sorts of semantics of "default" vs. "site-wide" etc.  That kind of thing is for the user-facing text.  For the underlying value in terms of the variable it seems that "default_" is extra noise.  There is the mode and then it can be overridden, but it's not that there is a default in the system and then different situations that lead to different modes.  I guess I could be convinced if you really feel strongly, but in terms of labelling I'll address those below with your other comment.

src/Form/AdminSettingsForm.php
eshtadc: OK - so I'm still very much not fond of the term "deployment" and I think using a "Content" heading would be even more confusing as everything in the "Data Collection" grouping appears to have more to do with content than the replacement mode does.  It is more of an Advanced setting because we will have a default value (currently untrusted as it is what we have tested most - but likely to change to trusted).  The sysadmin would only visit this section if they had a specific use-case to change the functionality for the whole site, hence "advanced configuration".  Overriding on a per-slot basis is extremely advanced use-case that really belongs more in documentation and less as an in-your-face additional possible layer of confusion.

There are definitely plans for descriptions, however, it was intended to add these in the d8 module refinement story as we have a better handle on use cases, etc. to include and/or (even better) a link to documentation.

@tottey and I had discussed this prior to my inclusion of the option in the admin (as the original story only requires it being set by drush config value).

config/install/acquia_lift.settings.yml
ynx: 1. The reason I feel "deployment" fits users' anticipation better than "replacement", is because when I was observing users, the impression I have perceived was they tend to think and ask "how do I put content here", and not what I as a dev would usually think "how do I replace this element by a content element". In other words "replacement" does not fit my user mentality. @tottey do you share the same thought or am I overthinking it? 
2. For me, this "extra" and/or "noise" isn't actually a bad thing, since it provides additional semantic value, and describes the nature of this parameter more accurately, leaving no ambiguity for the next dev. Plus, it is always a bonus for consistency, even just between the text and underlying parameter name. Lastly, there is no notable penalty for longer names, but redundant ones.

On the other hand, I am interested in showing you what I think for you consideration, and not trying to enforce - I am sure neither of us is terribly wrong, and if we are debating I'd prefer moving the story forward without change.

src/Form/AdminSettingsForm.php
eshtadc: FYI:  Added a sub-task regarding adding descriptions to the module: https://backlog.acquia.com/browse/LEX-1227

config/install/acquia_lift.settings.yml
tottey: I actually think replacement is the better approach as my observations with users is the opposite. Deployment carries with it a connotation of a heavy action and is somewhat technical - I'm "deploying" something or something is being "deployed." Most end-users I've talked to and tested very much understand the concept that the rules they are creating replace things. As such, my recommendation would be let's go with replacement.

src/Form/AdminSettingsForm.php
eshtadc: Addressed moving the setting to an advanced section here: https://github.com/acquia/acquia_lift/pull/238/commits/a498595714acdb674f1e616f0634e49a5eb72ff2

ynx: If @tottey is okay with it I am okay.

Commit looks great. Thanks!

PR #243, opened 2016-11-02
LEX-1519 Cache visitor_context values, if they are used in field / UD…

ynx:
…F data collection.

### DIFF
--- a/acquia_lift_profiles/acquia_lift_profiles.module
+++ b/acquia_lift_profiles/acquia_lift_profiles.module
@@ -216,6 +216,13 @@ function acquia_lift_profiles_page_build(&$page) {
   if (!empty($acquia_lift_profiles_field_mappings)) {
     $mappings['field'] = $acquia_lift_profiles_field_mappings;
   }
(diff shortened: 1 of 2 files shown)


acquia_lift_profiles/acquia_lift_profiles.module
```diff
@@ -216,6 +216,13 @@ function acquia_lift_profiles_page_build(&$page) {
   if (!empty($acquia_lift_profiles_field_mappings)) {
     $mappings['field'] = $acquia_lift_profiles_field_mappings;
   }
+  $cache_expiration = array();
+  foreach ($mappings as $mapping) {
+    foreach ($mapping as $mapped_context) {
+      $cache_expiration_key = 'visitor_context:' . str_replace('__', ':', $mapped_context);
+      $cache_expiration[$cache_expiration_key] = 'session';
+    }
+  }
```
ynx: Walk through all of field and UDF mappings, and set their bucket cache expiration to 'session'. This way, the values would get cached / stored session-wide, and available to data collection.

PR #245, opened 2016-11-15
CHMS-1012: Initial Drupal 7 offering for Lift 3

eshtadc:
The ultimate goal is feature parity with Drupal 8 module. This would become branch 7.x-3.x of the module. At this time it includes:
* Ability to set identity parameters
* Ability to map taxonomies
* Ability to map UDF (user defined) fields
* Configuration of Lift endpoints
* Rendering of values into meta tags similar to D8 rather than the drupalSettings settings

### DIFF
--- a/acquia_lift.admin.inc
+++ b/acquia_lift.admin.inc
@@ -1,0 +1,410 @@
+<?php
+
+/**
+ * @file acquia_lift.admin.inc
+ * Provides functions needed for the admin UI.
+ */
+
+
+function acquia_lift_admin_form($form, &$form_state) {
+  $is_configured = acquia_lift_is_configured();
+  $form['credentials'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('Acquia Lift credential'),
+    '#collapsible' => $is_configured,
+    '#collapsed' => $is_configured,
+    '#tree' => FALSE,
+  );
+  _acquia_lift_admin_build_credential_form($form['credentials'], $form_state);
+
+  $form['data_collection'] = array(
+    '#type' => 'vertical_tabs',
+    '#title' => t('Data collection settings'),
+  );
+
+  $form['identity'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('Identity'),
+    '#group' => 'data_collection',
+    '#collapsible' => TRUE,
+    '#tree' => FALSE,
+    '#weight' => 10,
+  );
+  _acquia_lift_admin_build_identity_form($form['identity'], $form_state);
+
+  $form['acquia_lift_field_mappings'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('Field mappings'),
+    '#description' => t('Create !taxonomy_link and map to "content section", "content keywords", and "persona" fields.', array(
+        '!taxonomy_link' => l(t('Taxonomy Vocabulary'), 'admin/structure/taxonomy'))
+    ),
+    '#group' => 'data_collection',
+    '#collapsible' => TRUE,
+    '#tree' => TRUE,
+    '#weight' => 20,
+  );
+  _acquia_lift_admin_build_field_mappings_form($form['acquia_lift_field_mappings'], $form_state);
+
+  $form['acquia_lift_udf_mappings'] = array(
+    '#tree' => TRUE,
+  );
+  $form['acquia_lift_udf_mappings']['person'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('User person mappings'),
+    '#description' => t('Map taxonomy terms to Visitor Profile person fields in Acquia Lift. Select a Taxonomy Reference Field that, if present, will map the value of the specified field to the Acquia Lift Profile for that specific visitor. No options available? Create !taxonomy_vocabulary_link and map the corresponding value.', array(
+        '!taxonomy_vocabulary_link' => l(t('Taxonomy Vocabulary'), 'admin/structure/taxonomy'))
+    ),
+    '#group' => 'data_collection',
+    '#collapsible' => TRUE,
+    '#weight' => 30,
+  );
+  _acquia_lift_admin_build_udf_mappings_form($form['acquia_lift_udf_mappings']['person'], $form_state, 'person');
+
+  $form['acquia_lift_udf_mappings']['touch'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('User touch mappings'),
+    '#description' => t('Map taxonomy terms to Visitor Profile touch fields in Acquia Lift. Select a Taxonomy Reference Field that, if present, will map the value of the specified field to the Acquia Lift Profile for that specific visitor. No options available? Create !taxonomy_vocabulary_link and map the corresponding value.', array(
+      '!taxonomy_vocabulary_link' => l(t('Taxonomy Vocabulary'), 'admin/structure/taxonomy'))
+    ),
+    '#group' => 'data_collection',
+    '#collapsible' => TRUE,
+    '#weight' => 40,
+  );
+  _acquia_lift_admin_build_udf_mappings_form($form['acquia_lift_udf_mappings']['touch'], $form_state, 'touch');
+
+  $form['acquia_lift_udf_mappings']['event'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('User event mappings'),
+    '#description' => t('Map taxonomy terms to Visitor Profile event fields in Acquia Lift. Select a Taxonomy Reference Field that, if present, will map the value of the specified field to the Acquia Lift Profile for that specific visitor. No options available? Create !taxonomy_vocabulary_link and map the corresponding value.', array(
+        '!taxonomy_vocabulary_link' => l(t('Taxonomy Vocabulary'), 'admin/structure/taxonomy'))
+    ),
+    '#group' => 'data_collection',
+    '#collapsible' => TRUE,
+    '#weight' => 50,
+  );
+  _acquia_lift_admin_build_udf_mappings_form($form['acquia_lift_udf_mappings']['event'], $form_state, 'event');
+
+  $form['visibility'] = array(
+    '#type' => 'fieldset',
+    '#description' => t('Lift will skip data collection on those URLs and their aliases.'),
+    '#title' => t('Visibility'),
+    '#group' => 'data_collection',
+    '#collapsible' => TRUE,
+    '#tree' => FALSE,
+    '#weight' => 60,
+  );
+  _acquia_lift_admin_build_visibility_form($form['visibility'], $form_state);
+
+  $form['advanced'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('Advanced configuration'),
+    '#collapsible' => TRUE,
+    '#collapsed' => TRUE,
+    '#tree' => FALSE,
+    '#weight' => 70,
+  );
+  _acquia_lift_admin_build_advanced_confirmation_form($form['advanced'], $form_state);
+  $form['submit'] = array(
+    '#type' => 'submit',
+    '#value' => t('Save configuration'),
+    '#weight' => 80,
+  );
+  return $form;
+}
+
+function _acquia_lift_admin_build_credential_form(&$form, &$form_state) {
+  $form['acquia_lift_account_id'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Account ID'),
+    '#description' => t('Your Lift subscription\'s account ID.'),
+    '#default_value' => variable_get('acquia_lift_account_id', ''),
+    '#required' => TRUE,
+  );
+  $form['acquia_lift_site_id'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Site ID'),
+    '#description' => t('Current site\'s site ID. WARNING: different sites must use different value here, even between a staging and a production of the same site.'),
+    '#default_value' => variable_get('acquia_lift_site_id', ''),
+    '#required' => TRUE,
+  );
+  $form['acquia_lift_assets_url'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Assets URL'),
+    '#description' => t('Your Lift application\'s assets URL.  It determines which version of the Lift application is being used.'),
+    '#field_prefix' => 'https://',
+    '#default_value' => _acquia_lift_admin_clean_url(variable_get('acquia_lift_assets_url', '')),
+    '#required' => TRUE,
+  );
+  $form['acquia_lift_decision_api_url'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Decision API URL'),
+    '#description' => t('Your Lift Decision API\'s URL.  Unless explicitly instructed, leave empty to use default URL.'),
+    '#field_prefix' => 'https://',
+    '#default_value' => _acquia_lift_admin_clean_url(variable_get('acquia_lift_decision_api_url', '')),
+  );
+  $form['acquia_lift_oauth_url'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Authentication URL'),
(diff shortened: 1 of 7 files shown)


acquia_lift.info
```diff
@@ -0,0 +1,9 @@
+name = Acquia Lift
+description = Provides integration to the Acquia Lift service for personalization and visitor data collection.
+configure = admin/config/services/acquia_lift
+
+package = Acquia
+core = 7.x
+
+files[] = tests/acquia_lift.test
+files[] = tests/acquia_lift_unit.test
```
eshtadc: Oops - I'll take this out.

acquia_lift.module
```diff
@@ -0,0 +1,262 @@
...
+
+/**
+ * Implements hook_menu().
+ */
+function acquia_lift_menu() {
+  $items = array();
+  $items['admin/config/services/acquia_lift'] = array(
+    'type' => MENU_NORMAL_ITEM,
+    'title' => 'Acquia Lift',
+    'description' => 'Configuration settings for the Acquia Lift integration module.',
```
ynx: Not a big deal, but wording "Configuration settings for the Acquia Lift integration module." could better just be "Configure Acquia Lift settings.". This is because:
- We no longer call it integration module.
- To match D8 wording.

acquia_lift.module
```diff
@@ -0,0 +1,262 @@
...
+ *
+ * @return array
+ *   An array of UDF types defined in Acquia Lift Profile Manager with keys
+ *   as the type name and values as the number allowed.
+ */
+function acquia_lift_get_udf_types() {
+  return array(
+    'person' => 50,
+    'touch' => 20,
+    'event' => 20,
```
ynx: Event count should be 50.

acquia_lift.admin.inc
```diff
@@ -0,0 +1,410 @@
...
+  variable_set('acquia_lift_oauth_url', $oauth_url);
+
+  // The rest of the variables are set as is.
+  $standard_variables = array(
+    'acquia_lift_identity_param',
+    'acquia_lift_identity_type_param',
+    'acquia_lift_default_identity_type',
+    'acquia_lift_account_id',
+    'acquia_lift_site_id',
+    'acquia_lift_ignore_path_patterns',
```
ynx, line 346: The reason why D8 named this variable "path_patterns" instead of "ignore_path_patterns" is because I observed a tendency/design on other frameworks that people also possibly want a blacklist/whitelist check box to switch this "path_patterns" to be a blacklist or whitelist pattern.

There is no business need for "whitelist" functionality for now, but to leave a bit room for future and skip a potential data migration hook, I have skipped the "ignore" in its D8 naming.

I don't have a firm stance on naming the variable either way, so this is just a note instead of a change request.

tests/acquia_lift.test
```diff
@@ -0,0 +1,561 @@
...
+      'acquia_lift_ignore_path_patterns',
+      'acquia_lift_content_replacement_mode',
+      'acquia_lift_field_mappings[content_section]',
+      'acquia_lift_field_mappings[content_keywords]',
+      'acquia_lift_field_mappings[persona]',
+    );
+    for ($i = 1; $i <= 50; $i++) {
+      $fields[] = 'acquia_lift_udf_mappings[person][person_udf' . $i . ']';
+    }
+    for ($i = 1; $i <= 20; $i++) {
```
ynx, line 97: Event UDF count should be 50.

tests/acquia_lift.test
```diff
@@ -0,0 +1,561 @@
...
+    $expected_tags = implode(',', $this->tags['two_tags']);
+    $this->assertLiftMetaTag('person_udf1', $expected_tags);
+    $this->assertNoLiftMetaTag('touch_udf20');
+    $this->assertLiftMetaTag('event_udf5', $expected_tags);
+  }
+
+  /**
+   * Helper function to configure acquia_lift settings.
+   */
+  protected function configureLift() {
```
ynx: Very minor - all of the "protected" functions in this file can better be "private". This is because AcquiaLiftsWebTest class is not meant to be extended, and we can just follow OOP's maximum encapsulation guideline.

acquia_lift.module
```diff
@@ -0,0 +1,262 @@
...
+  return $groups;
+}
+
+function _acquia_lift_add_meta_tag($name, $value) {
+  $tag = array(
+    '#type' => 'html_tag',
+    '#tag' => 'meta',
+    '#attributes' => array(
+      'content' =>  $value,
+      'itemprop' => 'acquia_lift:' . $name,
```
ynx, line 256: Not a big deal, but in D8, the "itemprop" is rendered before "content". I understand it's semantically the same, but it would be nice if we can keep the exact same behaviour, one way or another.

Notice if we are to switch line 256 and 257, some tests may have to be updated.

acquia_lift.context.inc
```diff
@@ -0,0 +1,156 @@
...
+  // Add the content type of the page if this is the primary node displayed.
+  if ($primary && empty($page_context['post_id'])) {
+    $page_context['content_type'] = $node->type;
+    $page_context['content_title'] = $node->title;
+    $page_context['published_date'] = $node->created;
+    $page_context['post_id'] = $node->nid;
+    $account = user_load($node->uid);
+    $page_context['author'] = $account->name;
+    $page_context['page_type'] = 'node page';
+    $page_context['type'] = $node->type;
```
ynx: I don't think we have a "type" page context. Do you mean "content_type"? However "content_type" is already defined at line 78.

acquia_lift.admin.inc
```diff
@@ -0,0 +1,410 @@
...
+    ));
+    $form[$udf_title] = $select;
+  }
+}
+
+function _acquia_lift_admin_build_visibility_form(&$form, &$form_state) {
+  $form['acquia_lift_ignore_path_patterns'] = array(
+    '#type' => 'textarea',
+    '#title' => t('Path patterns'),
+    '#default_value' => variable_get('acquia_lift_ignore_path_patterns', "/admin\n/admin/*\n/batch\n/node/add*\n/node/*/*\n/user/*/*\n/block/*"),
```
ynx: This is an error in D8 module - `/user/*/*\n` should be `/user/*\n`. This is fixed in LEX-1558.

Notice `/user/*\n` also blacklists `/user`, which is different than D8 but just what we want in D7.

acquia_lift.admin.inc
```diff
@@ -0,0 +1,410 @@
...
+      '@identity_parameter_display_value' => $identity_parameter_display_value,
+      '@identity_type_parameter_display_value' => $identity_type_parameter_display_value,
+      '@default_identity_type_default_value' => $default_identity_type_default_value,
+    ]),
+    '#default_value' => variable_get('acquia_lift_identity_type_param'),
+    '#states' => array(
+      'visible' => array(
+        ':input[name="acquia_lift_identity_param"]' => ['!value' => ''],
+      ),
+    ),
```
ynx, line 187: As per your recommendation, I have updated the D8 module to use the same '#states' for acquia_lift_default_identity_type. I think we can do that in here too.

acquia_lift.admin.inc
```diff
@@ -0,0 +1,410 @@
...
+    '#type' => 'textfield',
+    '#title' => t('Default Identity Type'),
+    '#description' => t('The identity type category to use by default. Leave this field blank to default to <ins>@default</ins>. Example: <ins>@default_identity_type_display_value</ins> is "?@identity_parameter_display_value=jdoe01&@identity_type_parameter_display_value=<ins>@default_identity_type_display_value</ins>", while blank is the same as "?@identity_parameter_display_value=jdoe01&@identity_type_parameter_display_value=<ins>@default</ins>"', [
+      '@default' => 'email',
+      '@identity_parameter_display_value' => $identity_parameter_display_value,
+      '@identity_type_parameter_display_value' => $identity_type_parameter_display_value,
+      '@default_identity_type_display_value' => $default_identity_type_display_value,
+    ]),
+    '#default_value' => variable_get('acquia_lift_default_identity_type'),
+    '#placeholder' => ACQUIA_LIFT_DEFAULT_IDENTITY_TYPE_DEFAULT,
```
ynx: "#placeholder" doesn't work in D7 natively. It was done in 7.x-2.x through a "[hack](https://github.com/acquia/acquia_lift/blob/7.x-2.x/acquia_lift_profiles/acquia_lift_profiles.admin.inc#L123-L125)".

acquia_lift.module
```diff
@@ -0,0 +1,262 @@
...
+  // If there are querystring parameters present corresponding to the configured
+  // identity params, we add the identity and identity type to the meta tags.
+  $identity_param = variable_get('acquia_lift_identity_param', '');
+  $identity_type_param = variable_get('acquia_lift_identity_type_param', '');
+  if (!empty($identity_param)) {
+    $all_params = drupal_get_query_parameters();
+    if (isset($all_params[$identity_param])) {
+      // Set the identity type to the default either as specified by the user or
+      // as the application default.
+      $identity_type = variable_get('acquia_lift_default_identity_type') ?: ACQUIA_LIFT_DEFAULT_IDENTITY_TYPE_DEFAULT;
```
ynx: How about: ```$identity_type = variable_get('acquia_lift_default_identity_type', ACQUIA_LIFT_DEFAULT_IDENTITY_TYPE_DEFAULT);```

Warning: not tested.

acquia_lift.module
```diff
@@ -0,0 +1,262 @@
...
+      if (!empty($identity_type_param) && isset($all_params[$identity_type_param])) {
```
ynx: I think `!empty` is better than `isset` here, because it guards malformed query string such as `?identity=shay1117a&identityType`, allowing the identityType to fallback to the default.

The D8 counter-part is updated in LEX-1558.

acquia_lift.admin.inc
```diff
@@ -0,0 +1,410 @@
...
+function _acquia_lift_admin_build_identity_form(&$form, &$form_state) {
+  $identity_parameter_display_value = variable_get('acquia_lift_identity_param', 'identity');
+  $identity_type_parameter_display_value = variable_get('acquia_lift_identity_type_param', 'identityType');
+  $default_identity_type_display_value = variable_get('acquia_lift_default_identity_type', ACQUIA_LIFT_DEFAULT_IDENTITY_TYPE_IDENTIFIER);
+  $default_identity_type_default_value = variable_get('acquia_lift_default_identity_type', ACQUIA_LIFT_DEFAULT_IDENTITY_TYPE_DEFAULT);
+
+  $form['acquia_lift_identity_param'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Identity Parameter'),
+    '#description' => t('The query string parameter for specific visitor information, such as an email address or social media username, which is sent to the Acquia Lift service. Example: ?<ins>@identity_parameter_display_value</ins>=jdoe01', [
```
ynx: @tottey and Steve finalized the wording. I have implemented the updates in 850cb825c38de31e5734cca80d73015d2dae9fcc. There is also a screenshot `description.png` in LEX-1578 to show how it looks in D8.

tests/acquia_lift.test
```diff
@@ -0,0 +1,561 @@
...
+   *
+   * @param array $autocreate
+   *   (optional) An array of term names to autocreate. Defaults to array().
+   * @param array $existing
+   *   (optional) An array of existing term IDs to add.
+   *
+   * @return object
+   *   The node object.
+   */
+  function createArticleWithTerms($autocreate = array(), $existing = array()) {
```
ynx: Missing protected/private here?

acquia_lift.module
eshtadc: I wrote my tests to work in either order but have to admit I was puzzled why they were written out of order.  Thanks for pointing it out.  Darn drupal.

acquia_lift.info
eshtadc: Done https://github.com/acquia/acquia_lift/pull/245/commits/9ceb7aca6c1891c2f0c6e62154d163063640999f

acquia_lift.module
eshtadc: I updated it here https://github.com/acquia/acquia_lift/pull/245/commits/c0f188b9f638dc248453dea92a05510e7f184ee4
This had been what was already in use in the D7 module (remember this was supposed to be a copy/paste story....).  We need to be a bit more consistent now when changes are made in one version to update both.  I believe the Drupal approach is to always start in the highest version and then backport, but internally it should just be part of the same story each time to include d7 when changes are made to d8.

acquia_lift.module
eshtadc: https://github.com/acquia/acquia_lift/pull/245/commits/8d35ed293af336a521d15ba95e8f745b366f8d82

acquia_lift.module
eshtadc: Really good catch :)
Are we doubly sure?  The D7 code always had it this way?  https://github.com/acquia/acquia_lift/blob/7.x-2.x/acquia_lift_profiles/acquia_lift_profiles.module#L757

acquia_lift.module
eshtadc: https://github.com/acquia/acquia_lift/pull/245/commits/6c8d181f0ae10f5352cfda60957d4191bd162565

acquia_lift.admin.inc
eshtadc: Parity is a moving target apparently.

eshtadc: https://github.com/acquia/acquia_lift/pull/245/commits/1e1462fd288f877e4422849c64b7fa4d60d0a42d

acquia_lift.admin.inc
eshtadc: https://github.com/acquia/acquia_lift/pull/245/commits/e48aec8e0063e4953a0ab6fa640c50b320ad786d

acquia_lift.admin.inc
eshtadc: https://github.com/acquia/acquia_lift/pull/245/commits/cc2877777dd7d5c386ea0c07f54a369fb721869d

tests/acquia_lift.test
eshtadc: Same as above - just want to double-check as the original/current d7 code has it at 20.  If this is true we need to update lift 2 also.

tests/acquia_lift.test
eshtadc: Not interested in getting into a discussion on this one for sure.  Things can always be changed when they need to.  https://github.com/acquia/acquia_lift/pull/245/commits/1bcf876abff2adc26a92f293e8c68927bf5b3696

tests/acquia_lift.test
eshtadc: https://github.com/acquia/acquia_lift/pull/245/commits/1bcf876abff2adc26a92f293e8c68927bf5b3696

acquia_lift.context.inc
eshtadc: It was taken from this code here: https://github.com/acquia/visitor_actions/blob/7.x-1.x/visitor_actions.module#L672

acquia_lift.admin.inc
eshtadc: Yeah - these crossed.  Since it got finalized while this was in code review I can go ahead and update this now.

eshtadc: https://github.com/acquia/acquia_lift/pull/245/commits/cdd6a4562d28c3bcde4e0dfc603f0369dbb0e151

acquia_lift.module
ynx: I talked to Steve. The 21 to 50 UDFs were introduced later. Lift should allow 50 event UDFs, so "50" is what we want here. 

I will update Lift 2.

tests/acquia_lift.test
ynx: Should be 50 (see last comment). I will update Lift 2.

acquia_lift.module
eshtadc: 👍

PR #247, opened 2016-11-17
LEX-1541 Remove the obsolete 'thumbnail URL' feature.

ynx:


.travis.yml
```diff
@@ -39,6 +39,7 @@ install:
 
   # Install drupal default profile and enable required modules.
   - /usr/bin/env PHP_OPTIONS="-d sendmail_path=`which true`" drush --verbose site-install --db-url=mysql://root:@127.0.0.1/drupal --yes
+  - composer install --dev
```
nickveenhof, line 42: Unrelated change?

config/schema/acquia_lift.settings.yml
```diff
@@ -98,19 +98,6 @@ acquia_lift.settings:
               type:
                 type: string
                 label: 'Item Type this is mapped to.'
-    thumbnail:
```
nickveenhof, line 101: I think there needs to be some upgrade function that removes this config no? What if it is populated? Will this work?

.travis.yml
ynx: The branch [failed on travis out of blue](https://travis-ci.org/acquia/acquia_lift/builds/176609525). Adding this line resolved the issue. I think we should keep it.

config/schema/acquia_lift.settings.yml
ynx: D8 continues to work with obsolete config variables, be it a full obsolete variable or a partial obsolete within a variable. It has happened a few times so I know this for sure now.

I thought about writing an update hook, but then, I thought why pest our customers for no benefit? So I didn't. =D

nickveenhof: Hmm, I don't fully agree here. We should not leave lingering config out there.

PR #248, opened 2016-11-24
LEX-1599 | Adding Acquia Lift Login Link to toolbar

nickveenhof:


assets/acquia_lift.js
```diff
@@ -0,0 +1,26 @@
...
+(function ($, Drupal) {
+
+    'use strict';
+
+    /**
+     * Attaches the JS test behavior to to weight div.
+     */
+    Drupal.behaviors.acquiaLiftOpenExperienceBuilder = {
+        attach: function (context, settings) {
+            $('.openLiftLink').click(function() {
```
wimleers: `.click()` is deprecated in jQuery. Use `.on('click')`, just like Drupal core does. Also, why jQuery? You can do this without.

assets/acquia_lift.js
```diff
@@ -0,0 +1,26 @@
...
+                if (window.AcquiaLiftPublicApi) {
+                    window.AcquiaLiftPublicApi.activate();
+                } else {
+                    alert('Sorry, Acquia Lift could not be found');
```
wimleers: How is this case possible? Why is this not a dependency of the asset library (where you're currently adding a dependency on `core/jquery`)? See `hook_library_info_build()` for how to define dynamic libraries, that allow you to create an asset library outside of `MODULENAME.libraries.yml`, and can include e.g. an asset URL that's entered by the user.

Also, why do you have asset URLs entered by users? Why is it not an API key?

acquia_lift.module
```diff
@@ -39,3 +40,56 @@ function acquia_lift_page_attachments(array &$page) {
...
+/**
+ * Implements hook_toolbar().
+ */
+function acquia_lift_toolbar() {
+  $links = array();
+  // Add a link to enable Lift 3.
+  $links['login_to_acquia_lift'] = array(
+    'title' => t('Login'),
+    // This will make the href attribute empty.
+    'url' => Url::fromUserInput('#'),
```
wimleers: Using a `href="#"` URL is a bad practice in web dev. This is not linking anywhere. That's a sure sign you _should not use a link_.

This is triggering some JS UI. So, use a button instead. Look at `contextual_toolbar()` for an example of this in core (and `contextual.toolbar.js` is the associated JS).

assets/acquia_lift.js
wimleers: Also, it's a best practice to have the actual logic in separate functions, for maintainability reasons.

So, all feedback combined, I'd propose something like

```
/**
 * Opens the Acquia Lift Experience Builder, if it is available.
 */
function openExperienceBuilder() {
  if (window.AcquiaLiftPublicApi) {
    window.AcquiaLiftPublicApi.activate();
  }
  else {
    alert('Sorry, Acquia Lift could not be found');
  }
}

Drupal.behaviors.acquiaLiftOpenExperienceBuilder = {
  attach: function (context, settings) {
    var trigger = document.querySelector('.toolbar .toolbar-bar .openLiftLink');
    trigger.addEventListener('click', openExperienceBuilder);
  }
};
```

wimleers: Ugh, this is going to add an event listener for every AJAX callback, because behaviors are attached for every single one of those. So you'll actually want:

```
Drupal.behaviors.acquiaLiftOpenExperienceBuilder = {
  attach: function (context, settings) {
    $(context)
        .find('.toolbar .toolbar-bar .openLiftLink')
        .once('acquia-lift-open-experience-builder')
        .on('click', openExperienceBuilder);
  }
};
```

(This is also why it's great to have the actual logic separate. It doesn't need to change now.)

wimleers: This requires `core/jquery.once` as a dependency BTW.

acquia_lift.module
```diff
@@ -12,6 +12,7 @@
  */
 
 use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Url;
```
wimleers: Unused now.

acquia_lift.module
```diff
@@ -39,3 +40,48 @@ function acquia_lift_page_attachments(array &$page) {
...
+  $items = [];
+  $items['acquia_lift'] = [
+    '#cache' => [
+      'contexts' => [
+        'user.permissions',
+      ],
+    ],
+  ];
+
+  if (!\Drupal::currentUser()->hasPermission('access acquia lift links')) {
```
wimleers: Nice, you thought about access control and cacheability metadata! Woot! :)

Tip for making this simpler: remove lines 49 through 59, and instead add this to lines 61–84:

```
  '#access' => AccessResult::allowedIfHasPermission('access acquia lift links'),
```

That looks easier on the eyes, doesn't it? Declarative code FTW.

assets/acquia_lift.js
```diff
@@ -0,0 +1,34 @@
...
+        if (window.AcquiaLiftPublicApi) {
+            window.AcquiaLiftPublicApi.activate();
+        }
+        else {
+            alert('Sorry, Acquia Lift could not be found');
+        }
+    }
+
+    /**
+     * Attaches the JS test behavior to to weight div.
```
wimleers: Wrong comment.

wimleers (approved):
lgtm :)

acquia_lift.libraries.yml
```diff
@@ -2,3 +2,7 @@ acquia-lift:
   css:
     state:
       assets/acquia_lift.css: {}
+  js:
+    assets/acquia_lift.js: {}
+  dependencies:
+    - core/jquery
```
wimleers: Oops, this still needs to become `core/jquery.once`.

It's only working by accident now, because something else is depending on that asset library.

acquia_lift.permissions.yml
```diff
@@ -1,3 +1,5 @@
 administer acquia lift:
   title: Administer Acquia Lift
   description: Administer settings of Acquia Lift module.
+access acquia lift links:
+  title: 'See the Acquia Lift links in the toolbar or elsewhere.'
```
ynx, line 5: Do we need a new permission for "access acquia lift links"? I couldn't think of a business case where an site admin can "access acquia lift links" but don't "administer acquia lift".

acquia_lift.module
```diff
@@ -39,3 +38,38 @@ function acquia_lift_page_attachments(array &$page) {
...
+    ],
+    '#access' => AccessResult::allowedIfHasPermission(\Drupal::currentUser(), 'access acquia lift links'),
+    '#wrapper_attributes' => [
+      'class' => ['acquia-lift-toolbar-tab'],
+    ],
+    '#attached' => [
+      'library' => [
+        'acquia_lift/acquia-lift',
+      ],
+    ],
```
ynx, line 71: Bizarrely, I don't think "#access" hides the "#attached" here, although I think that was the intention. You can go to the site front page in incognito without login, and see acquia_lift.js load; worse yet, acquia_lift.js is currently giving a JS error there for not finding "Drupal" object.

assets/acquia_lift.js
```diff
@@ -0,0 +1,35 @@
...
+  /**
+   * Opens the Acquia Lift Experience Builder, if it is available.
+   */
+  function openExperienceBuilder() {
+    if (window.AcquiaLiftPublicApi) {
+      window.AcquiaLiftPublicApi.activate();
+    }
+    else {
+      alert('Sorry, Acquia Lift could not be found');
+    }
```
ynx, line 19: I'd recommend that we use the same wording as Ember app's for consistency. I'd propose:

```Acquia Lift Experience Builder could not be loaded. Please check Lift admin page and make sure your account credentials are already set.  If this problem persists, please contact the Acquia Lift Team for support.```

File: acquia_lift.module [hunk 11]
```diff
@@ -39,3 +38,38 @@ function acquia_lift_page_attachments(array &$page) {
...
+  ];
+
+  return $items;
+}
```
ynx, line 75: The following would be how I would have approached this:

* Create Drupal\acquia_lift\Service\Helper\ShortcutHelper class, just like the HelpMessageHelper class.
* Move all logic into there.
* Unit test the Helper.
* Simpletest / functionally test the button exists.

This is to follow MVC's "fat-model, skinny-controller" guideline.

This note is FYI only, I wouldn't demand other developers to do the same - I'll be okay if you skip this note in partial or full.

acquia_lift.module
```diff
@@ -39,3 +38,38 @@ function acquia_lift_page_attachments(array &$page) {
...
+      '#attributes' => [
+        'title' => t('Acquia Lift'),
+        'class' => ['toolbar-icon', 'toolbar-icon-acquia-lift'],
+        // This particular id will make sure the JS behavior is attached to the
+        // link.
+        'id' => ['openLiftLink'],
+        'aria-pressed' => 'false',
+      ],
+    ],
+    '#access' => AccessResult::allowedIfHasPermission(\Drupal::currentUser(), 'access acquia lift links'),
```
ynx, line 63: I think we also need to ```&& $path_context->shouldAttach()```. This is because shouldAttach() conditionally stops Experience Builder from loading on pages, then on those pages you'd also not wanting to see the "Acquia Lift" shortcut. In this way, we'd never reach ```alert('Sorry, Acquia Lift could not be found');```.

acquia_lift.permissions.yml
nickveenhof: I actually do think there is a need for this. I can think of content editors seeing the link while not being able to change the configuration of the module.

acquia_lift.module
nickveenhof: I am not following what you want here or what that shouldAttach is doing. Can you explain? Apologies!

acquia_lift.module
nickveenhof: @wimleers , can you help us out here?

acquia_lift.module
nickveenhof: Wow, no - I completely resists. It goes way beyond the scope of this story which is just to add a simple link. Adding tests is fine for me, but I don't really see the added value of the further abstractions. Adding an integration test that simply verifies that the link is there and if you click it it shows you the error should be more than sufficient.

assets/acquia_lift.js
nickveenhof: I don't have a hard opinion. The text came from the link I already had. But sure, we can change it.

acquia_lift.module
ynx: No problem. PathContext service's shouldAttach() currently checks two things: 1) does Lift has good credential? 2) is current page URL suitable for loading Lift? If yes to both, it loads lift.js (and everything else). (Please see acquia_lift_page_attachments() in acquia_lift.module)

I think we should use shouldAttach() here as well, because the shortcut only works if we know lift.js actually loads on the page.

PR #249, opened 2016-12-06
Adding JS dependency for the toolbar JS and make sure that it only lo…

nickveenhof:
…ads when the toolbar is loaded


acquia_lift.module
```diff
@@ -35,14 +35,17 @@ function acquia_lift_page_attachments(array &$page) {
   $path_context->populate($page);
   $page_context->populate($page);
 
-  // Attach lift CSS to all pages.
-  $page['#attached']['library'][] = 'acquia_lift/acquia-lift';
+  // Attach lift CSS to all pages if the toolbar is present.
```
nickveenhof: Curious if we should rename the acquia_lift library to acquia_lift_toolbar, as this has only changes (also the CSS) to the toolbar stuff.

nickveenhof:
@ynx , changed some of the code a bit to make it better and to fix that bug that you were seeing. Because it is using that hooks system, this isn't the most optimal way of doing things. The hook does not respect render array caching from the #access call it seems. I tried adding cache contexts but that didn't help either. This is the best I can come up with

acquia_lift.module
```diff
@@ -35,15 +35,25 @@ function acquia_lift_page_attachments(array &$page) {
...
 
 /**
  * Implements hook_toolbar().
  */
 function acquia_lift_toolbar() {
+  // Create and attach settings and library only when path context agrees on attaching.
+  $path_context = \Drupal::service('acquia_lift.service.context.path_context');
   $items = [];
+  // If this link should not be supported
+  if (!$path_context->shouldAttach()) {
```
ynx: Could this better be ```!$path_context->shouldAttach() || !\Drupal::currentUser()->hasPermission('access acquia lift links')```?

acquia_lift.module
```diff
@@ -60,16 +70,20 @@ function acquia_lift_toolbar() {
         'aria-pressed' => 'false',
       ],
     ],
-    '#access' => AccessResult::allowedIfHasPermission(\Drupal::currentUser(), 'access acquia lift links'),
+    '#access' => \Drupal::currentUser()->hasPermission('access acquia lift links'),
+    '#cache' => [
+      'keys' => ['toolbar'],
+      'contexts' => ['user.permissions'],
```
ynx, line 77: I was challenged by cache context before. Instead of "user.permissions", have you tried just "user" or "user.permissions:access acquia lift links"?

nickveenhof:
Resolved the cache issues. Let's hope simpletest agrees

PR #251, opened 2016-12-09
fix missing test

nickveenhof:


src/Service/Context/PathContext.php
```diff
@@ -178,4 +185,25 @@ private function setContextIdentity($identity, $identityType) {
     $this->htmlHeadContexts['identity:' . $sanitized_identity_type] = $sanitized_identity;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheMaxAge() {
+    return Cache::PERMANENT;
```
wimleers: ```return $this->settings->getCacheMaxAge()```

src/Service/Context/PathContext.php
```diff
@@ -178,4 +185,25 @@ private function setContextIdentity($identity, $identityType) {
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheContexts() {
+    return [];
```
wimleers: ```return $this->settings->getCacheContexts()```

src/Service/Context/PathContext.php
```diff
@@ -11,7 +14,14 @@
 use Drupal\acquia_lift\Service\Helper\PathMatcher;
 use Drupal\acquia_lift\Service\Helper\SettingsHelper;
 
-class PathContext extends BaseContext {
+class PathContext extends BaseContext implements CacheableDependencyInterface {
+
+  /**
+   * Acquia Lift settings.
+   *
+   * @var ImmutableConfig
```
wimleers: Needs FQCN.

src/Service/Context/PathContext.php
wimleers: Also, shouldn't `PathContext` have a cache context, like `url.path`?

PR #252, opened 2016-12-18
Add Lift Inspector to D8  Module

briancwald:
I have my Lift Inspector from Demo Framework to go in Lift D8 module, including a number of reactors to make it work generically across all accounts.

## To use: 
- Go to lift config page, turn on Lift Inspector
- In the browser with Lift enabled, press "Ctrl + i"

## New features:
**- Added inspector as optional (off by default) feature that can be enabled and disabled from the Acquia Lift config page**
screenshot 2016-12-18 16 08 55

**- Style Inspector to closer match Lift sidebar styles**
screenshot 2016-12-18 16 03 35

**Deep linking to segments, customer and current users tracking ID (note, must set the subdomain for lift profile manager in the settings page)**

## Fixes
- Better key mapping so that you can prompt the inspector from any browser.
- Callback on dialog load to lock the dialog from background scrolling when open


nickveenhof:
acquia_lift.settings:advanced.inspector_status missing schema,
    acquia_lift.settings:advanced.profile_manager_url missing schema in

So you need to adjust the config files :)

acquia_lift.module
```diff
@@ -41,6 +41,17 @@ function acquia_lift_page_attachments(array &$page) {
...
     $page['#attached']['library'][] = 'acquia_lift/acquia-lift-toolbar';
   }
+
+  // Get the config settings to see if the lift inspector is enabled
+  $config = \Drupal::config('acquia_lift.settings');
+  $lift_inspector = $config->get('advanced.inspector_status');
+  $profile_manager_url = $config->get('advanced.profile_manager_url');
+
+  // If the lift inspector is enabled, attach the library and pass the profile manager URL to drupalSettings
+  if ($lift_inspector) {
```
nickveenhof: Better to set === TRUE if we are talking about a boolean. As PHP can be quite quirky sometimes when not strictly defined.

assets/lift_inspector/acquia_lift_inspector.css
```diff
@@ -0,0 +1,182 @@
...
+}
+
+#lift-inspector .lift-inspector-wrapper-fullwidth {
+    width: 98%;
+}
+
+#lift-inspector .lift-inspector-wrapper-fullwidth .container div {
+    display: inline-block;
+}
+
```
nickveenhof, line 173: Remove double newlines

assets/lift_inspector/acquia_lift_inspector.js
```diff
@@ -0,0 +1,125 @@
+/**
+ * @file acquia_lift_inspector.js
+ */
+(function ($, Drupal) {
+  'use strict';
+
+  Drupal.behaviors.acquia_lift = {
+    attach: function (context, settings) {
+
+      function show_inspector_modal() {
```
nickveenhof: Let's make this camelcase, as good JS citizens

assets/lift_inspector/acquia_lift_inspector.js
```diff
@@ -0,0 +1,125 @@
...
+        var $modal_element, modal_template;
+
+        $('#lift-inspector').remove();
+
+        //Get the Profile Manager URL from the config settings
+        var accountSubDomain = drupalSettings.acquia_lift.acquia_lift_inspector.profile_manager_url;
+
+        if (accountSubDomain) {
```
nickveenhof, line 18: is the profile_manager_url always only a prefix?

assets/lift_inspector/acquia_lift_inspector.js
```diff
@@ -0,0 +1,125 @@
...
+          var accountURL = 'https://' + accountSubDomain + '.lift.acquia.com';
+        } else {
+          var accountURL = 'https://lift.acquia.com';
+        };
+
+        modal_template = _.template(
+          '<div id="lift-inspector">' +
```
nickveenhof, line 25: Can't we somehow load this in from somewhere else? html in JS is so ugly :(

assets/lift_inspector/acquia_lift_inspector.js
```diff
@@ -0,0 +1,125 @@
...
+          beforeClose: function(event, ui) {$(dialogContainerSelector).removeClass(dialogActiveClassName);}
+        }).showModal();
+      }
+
+      // $(document).ready(function(){
+      //   show_inspector_modal();
+      // })
+
+
+      $(document).keypress(function (e) {
```
nickveenhof, line 117: Will this work in all browsers?

src/Form/AdminSettingsForm.php
```diff
@@ -387,6 +387,21 @@ private function buildAdvancedConfigurationForm() {
       '#options' => ['trusted' => t('Trusted'), 'untrusted' => t('Untrusted')],
     ];
 
+    $form['inspector_status'] = [
+      '#type' => 'checkbox',
+      '#title' => t('Enable Lift Inspector'),
+      '#default_value' => $settings['inspector_status'],
+      '#description' => t('Enabling this feature will allow you to use type "cntl + i" to get lift information about the session'),
+    ];
+    $form['profile_manager_url'] = [
```
nickveenhof, line 396: Are you sure we can't "guess" this instead of asking the user to fill in this setting? I'll ask the team and get back to you on this.

nickveenhof: Also, we do not prevent them here from filling in the complete url. To reduce mistakes I'd really want us to remove this from the patch/module and fill it in automatically

nickveenhof (changes requested):

nickveenhof:
Also I'd really like some Drupal 8 JS tests now that those are available before we get this in. See the facets module for examples on how that works.

assets/lift_inspector/acquia_lift_inspector.js
briancwald: Yes, I tested it in most browsers besides IE, but according to the docs, its supported.

src/Form/AdminSettingsForm.php
briancwald: I agree, I would have much preferred to get that information directly from the JS but it is not available and this was the only method available to me. I didn't want that to stop me from making progress.

assets/lift_inspector/acquia_lift_inspector.js
briancwald: AFAICT, but we should ask Steve. In the case there is no prefix, I have accounted for that in the docs.

nickveenhof:
Closing as duplicate of https://github.com/acquia/acquia_lift/pull/253. That PR is a refactored version

PR #257, opened 2017-06-23
LAT-548 | Support page view UUID.

ynx:


tests/acquia_lift.test
```diff
@@ -439,6 +439,35 @@ class AcquiaLiftsWebTest extends DrupalWebTestCase {
     $this->assertLiftMetaTag('event_udf5', $expected_tags);
   }
 
+  function testNonNodePage() {
```
ynx, line 442: Back filling some tests.

PR #260, opened 2017-08-31
LAT-895 | Exclude lift.js from customer site's JS aggregation.

ynx:


acquia_lift.module
```diff
@@ -140,9 +140,10 @@ function acquia_lift_page_build(&$page) {
   // Add the lift js bootstrap loader
   $js_file = $assets_url . '/lift.js';
   $attached['js'] = array(
-    'type' => 'file',
-    'file' => $js_file,
-    'weight' => JS_DEFAULT - 10,
+    $js_file => array(
+      'type' => 'external',
```
abarriosr, line 144: Is $js_file is an external file? You could also use the options "preprocess" = FALSE as explained here https://api.drupal.org/api/drupal/includes%21common.inc/function/drupal_add_js/7.x

ynx: Setting the file type to "external" (as opposed to "file") automatically excludes the file being preprocessed. Also, the lift.js is indeed an "external" anyway.

ynx: $js_file value usually looks like ```https://lift3assets.lift.acquia.com/stable/lift.js```.

chilic (approved):

acquia_lift.module
chilic: @ynx In this case, what about set default value here
`  $assets_url = variable_get('acquia_lift_assets_url');`
?

ynx: @chilic Assets URL field is required on the module's configuration form; also at the start of this function, it calls to check acquia_lift_is_configured() which ensures acquia_lift_assets_url is already set, otherwise early exit.

matthewmorley: The only things that struck me as odd was the $js_file as a key, but only because I dont often do that I guess... vs: 

$attached['js'][$js_file] = array(
  'type' => 'external',
  'weight' => 0 - 10,
);

Looks good, no change required.

ynx:
@abarriosr, @chilic, and @matthewmorley thank you for the input! Usually the reviewer manually tests the code, but in this case I feel this change is small and I have manually tested extensive enough. Merged.

acquia_lift.module
ynx: @matthewmorley Right! Yours is indeed more sleek.

ynx: @matthewmorley I have updated 7.x-3.x as you mentioned: 9e96050bbf2320043df4712b014c6f363a095e67

PR #262, opened 2017-11-09
LAT-923 | 'Content replacement mode' to support one more option, 'customized'.

ynx:


ynx:
@chuckcarpenter If you could take a look at this as well ... (D7)

chuckcarpenter (approved):

PR #267, opened 2018-08-24
Add bootstrap modes to D7 for manual loading of Lift personalize method

kaynenh:


acquia_lift.admin.inc
```diff
@@ -256,6 +256,18 @@ function _acquia_lift_admin_build_visibility_form(&$form, &$form_state) {
 }
 
 function _acquia_lift_admin_build_advanced_confirmation_form(&$form, &$form_state) {
+  $form['acquia_lift_bootstrap_mode'] = [
+    '#type' => 'radios',
+    '#title' => t('Bootstrap mode'),
+    '#description' => t('How would you like to bootstrap Lift? Setting manul allows you to run additional scripts (useful if collecting and sending additional data or forcing an opt-in for users) but you must <a href="@docs-link" target="_blank">run the personalize() method in code</a> or Lift will not bootstrap.', array(
+      '@docs-link' => 'https://docs.acquia.com/lift/api/javascript/personalize/',
+    )),
```
ynx: Please update description to:
```
"Auto" means Lift scripts will automatically bootstrap and act as quickly as possible. "Manual" means Lift scripts will load but withhold itself from collecting data, delivering content, and allowing admins to login; this option is useful when you want to do things on your site (e.g. check a cookie, set field value) before you want Lift to start bootstrapping; to resume Lift's bootstrapping process, call AcquiaLiftPublicApi.personalize().
```

ynx: Again @tottey I hope my proposed description makes sense. Please update as needed.

acquia_lift.admin.inc
```diff
@@ -256,6 +256,18 @@ function _acquia_lift_admin_build_visibility_form(&$form, &$form_state) {
 }
 
 function _acquia_lift_admin_build_advanced_confirmation_form(&$form, &$form_state) {
+  $form['acquia_lift_bootstrap_mode'] = [
+    '#type' => 'radios',
+    '#title' => t('Bootstrap mode'),
+    '#description' => t('How would you like to bootstrap Lift? Setting manul allows you to run additional scripts (useful if collecting and sending additional data or forcing an opt-in for users) but you must <a href="@docs-link" target="_blank">run the personalize() method in code</a> or Lift will not bootstrap.', array(
+      '@docs-link' => 'https://docs.acquia.com/lift/api/javascript/personalize/',
```
ynx: No need for @docs-link now.

acquia_lift.module
```diff
@@ -109,6 +109,7 @@ function acquia_lift_page_build(&$page) {
     _acquia_lift_add_meta_tag('authEndpoint', variable_get('acquia_lift_oauth_url'));
   }
   _acquia_lift_add_meta_tag('contentReplacementMode', variable_get('acquia_lift_content_replacement_mode', 'untrusted'));
+  _acquia_lift_add_meta_tag('bootstrapMode', variable_get('acquia_lift_bootstrap_mode', ''));
```
ynx: Please rotate line 111 and 112 to keep the same order between bootstrap_mode and content_replacement_mode across the rest of the applications.

kaynenh: Should we default this to 'auto' here? Similar to defaulting content replacement to 'untrusted'?

chuckcarpenter: yes, we need to make sure that behavior is the same as current by default. Think of it as opt out.

tests/acquia_lift.test
```diff
@@ -131,6 +133,13 @@ class AcquiaLiftsWebTest extends DrupalWebTestCase {
     $this->assertText($submit_confirm_text);
     $this->assertEqual(variable_get('acquia_lift_ignore_path_patterns'), $test_path_patterns, 'Visibility path patterns were updated.');
 
+    // Test bootstrap mode settings
+    $this->assertEqual(variable_get('acquia_lift_bootstrap_mode'), 'auto', 'Bootstrap mode initially saved as auto mode.');
+    $edit['acquia_lift_bootstrap_mode'] = 'manual';
+    $this->drupalPost(NULL, $edit, $submit_button_text);
+    $this->assertText($submit_confirm_text);
+    $this->assertEqual(variable_get('acquia_lift_bootstrap_mode'), 'manual', 'Content replacement mode was updated to manual.');
```
ynx: "Content replacement mode" should be "Bootstrap mode".

ynx:
I just have also realized that if we search for "contentReplacementMode", there are a bunch of "bootstrapMode" lines that we would need to have. The same happens for D8 module.

PR #271, opened 2018-09-10
LEB-1917 | Make 8.x-3.x backwards compatible to Drupal 8.4.*.

ynx:
Drupal wants to use `drupal_set_message()` until `8.5` when `MessengerInterface` and `Messenger` service is made available. This PR is to make sure we fallback to `drupal_set_message()` if `Messenger` service is not available.

### DIFF
--- a/src/Form/AdminSettingsForm.php
+++ b/src/Form/AdminSettingsForm.php
@@ -52,10 +52,10 @@ class AdminSettingsForm extends ConfigFormBase {
    *
    * @param \Drupal\Core\Entity\EntityManagerInterface $entity_manager
    *   The entity manager.
-   * @param \Drupal\Core\Messenger\MessengerInterface $messenger
-   *   The messenger service.
+   * @param \Drupal\Core\Messenger\MessengerInterface|null $messenger
+   *   The messenger service (or null).
    */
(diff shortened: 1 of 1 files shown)


src/Form/AdminSettingsForm.php
```diff
@@ -64,9 +64,13 @@ public function __construct(EntityManagerInterface $entity_manager, MessengerInt
    * {@inheritdoc}
    */
   public static function create(ContainerInterface $container) {
+    // MessengerInterface was introduced by Drupal 8.5.
+    // This code is for backwards-compatibility to 8.4 and below.
+    $messenger = $container->has('messenger') ? $container->get('messenger') : null;
+
     return new static(
       $container->get('entity.manager'),
```
abarriosr, line 72: We should try not to use EntityManager either as it is deprecated and will be removed before D9.0, which is somewhere in the 8.x lifecycle.

ynx: I agree the work needs to be done. I ran out of bandwidth so I'd recommend that we implement the update in a later iteration of update.

abarriosr (approved):

ynx:
Thank you @abarriosr !

PR #278, opened 2019-05-01
Update ORCA integration

TravisCarden:
The latest version of ORCA includes many new features and improvements.


composer.json
```diff
@@ -1,12 +1,11 @@
...
-  "license": "GPL-2.0+",
-  "repositories": {
-    "drupal": {
-      "type": "composer",
-      "url": "https://packages.drupal.org/8"
+  "description": "Provides integration to the Acquia Lift service for personalization and visitor data collection.",
+  "license": "GPL-2.0-or-later",
+  "extra": {
+    "branch-alias": {
+      "dev-8.x-3.x": "3.x-dev"
```
scor, line 8: what is this line for? is it necessary? is it for ORCA? @TravisCarden

scor: I merged the PR but just curious....

TravisCarden: [Composer branch aliases](https://getcomposer.org/doc/articles/aliases.md#branch-alias) tell Composer to treat one version as equivalent to another one. This particular one helps compensate for Drupal's non-semantic version numbering scheme and prevents issues Composer sometimes has identifying installation candidates in pull requests. This line has broader application than just for ORCA, but ORCA does create most of the scenarios where it's important.

PR #280, opened 2019-05-07
LEB-2661 | Create new 8.x-4.x Lift Drupal Module

danikalmar:
- Cut a new version of the Drupal module using 8.x-3.10 as the baseline
- Remove "Authentication URL" field from the module credentials. This is no longer needed in v4
- Remove "Untrusted" mode option from the "Content replacement mode" selection in "Advanced Configuration"
- Remove "Content Hub Origin Site UUID" field and functionality from the module


scor:
Let's remove PHP 5 from travis.yml since it will always fail. Also, since Drupal core now requires PHP 7, I think we could also require PHP 7 in our module so that we don't have to maintain the 4.x branch to work with PHP 5.

scor:
the 4.x branch is missing the commit 45b2be95bee7fdeaac4457239263161b880f9810 you can compare between https://github.com/acquia/acquia_lift/commits/8.x-4.x and https://github.com/acquia/acquia_lift/commits/8.x-3.x

you may have to rebase your branch fixing the 4.x branch.

PR #285, opened 2019-07-04
LEB-2062 | Implement cdf version flag in the Lift Drupal module

danikalmar:


scor:
@danikalmar I think you're missing the drop down that is mentioned in https://backlog.acquia.com/browse/LEB-2062

danikalmar:
@scor you are probably right. I though I only had to implement one of the solutions.

scor (approved):

PR #288, opened 2019-12-13
LCH-3463 / LEB-4297 | Publish only required entities.

saxumVermes:


modules/acquia_lift_publisher/src/EventSubscriber/Publish/PublishOnlyRendered.php
```diff
@@ -0,0 +1,165 @@
...
+ * @package Drupal\acquia_lift_publisher\PruneCdf
+ */
+class PublishOnlyRendered implements EventSubscriberInterface {
+
+  /**
+   * Content Hub common actions.
+   *
+   * @var \Drupal\acquia_contenthub\ContentHubCommonActions
+   */
+  protected $commonActions;
```
ynx: Please change all `protected` functions and variables to `private`, unless they are intended for inheritance.

I understand Content Cloud projects use minimally `protected`. `acquia_lift` modules obey the universal standards of [Information Hiding](https://en.wikipedia.org/wiki/Information_hiding).

modules/acquia_lift_publisher/acquia_lift_publisher.links.task.yml
```diff
@@ -2,3 +2,8 @@ acquia_lift_publisher.entity_config_form:
   route_name: acquia_lift_publisher.entity_config_form
   title: 'Setup'
   base_route: acquia_lift_publisher.entity_config_form
+acquia_lift_publisher.settings_form:
+  route_name: acquia_lift_publisher.general_settings
+  title: 'Settings'
+  weight: 50
+  base_route: acquia_lift_publisher.entity_config_form
```
ynx: The new general settings form looks really nice, but I think it is an overkill. Also, logically I think it is still a part of the entity configuration.

Let's move the checkbox to the current Setup page, and eliminate the new tab.

modules/acquia_lift_publisher/src/EventSubscriber/Publish/PublishOnlyRendered.php
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    $entity = $event->getEntity();
+    $cdfs = $this->commonActions->getEntityCdf($entity);
+    foreach ($cdfs as $cdf) {
+      if ($cdf->getType() !== 'rendered_entity') {
+        continue;
+      }
+
+      $se_uuid = $this->getCdfEntityAttributeValue($cdf, 'source_entity');
```
kaynenh: Maybe $source_uuid here instead of $se_uuid just so it's more readable.

modules/acquia_lift_publisher/acquia_lift_publisher.module
```diff
@@ -0,0 +1,51 @@
...
+function acquia_lift_publisher_entity_insert(EntityInterface $entity) {
+  _acquia_lift_publisher_trigger_queue($entity);
+}
+
+/**
+ * Implements hook_entity_update().
+ */
+function acquia_lift_publisher_entity_update(EntityInterface $entity) {
+  _acquia_lift_publisher_trigger_queue($entity);
+}
```
ynx, line 26: I see insert and update. What about delete?

modules/acquia_lift_publisher/src/EventSubscriber/Publish/PublishOnlyRendered.php
```diff
@@ -0,0 +1,165 @@
...
+   */
+  public function onEnqueueCandidateEntity(ContentHubEntityEligibilityEvent $event): void {
+    if (!$this->personalizedContentPushIsActive()) {
+      return;
+    }
+
+    $entity = $event->getEntity();
+    $cdfs = $this->commonActions->getEntityCdf($entity);
+    foreach ($cdfs as $cdf) {
+      if ($cdf->getType() !== 'rendered_entity') {
```
kaynenh: We should double-check that this is enough. @ynx , do you know if this will fulfill the taxonomy terms needed for content recommendations?

modules/acquia_lift_publisher/acquia_lift_publisher.module
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Triggers the Content Hub export process.
+ *
+ * @param \Drupal\Core\Entity\EntityInterface $entity
+ *   The current entity.
+ */
+function _acquia_lift_publisher_trigger_queue(EntityInterface $entity): void {
```
ynx, line 34: This function is better suited in a separate class file in the `src/EventSubscriber/Publish` directory. `acquia_lift` modules attempts to keep the `.module` files as small as possible.

modules/acquia_lift_publisher/src/EventSubscriber/Publish/PublishOnlyRendered.php
```diff
@@ -0,0 +1,165 @@
...
+
+    $se_uuid = $this->getCdfEntityAttributeValue($rendered_entity, 'source_entity');
+    if (!$se_uuid) {
+      return;
+    }
+
+    $cdfs_to_keep = [$se_uuid];
+    $source_entity = $cdfs_to_remove[$se_uuid];
+    $tags = $this->getCdfEntityAttributeValue($source_entity, 'tags');
+    if (is_array($tags)) {
```
kaynenh, line 131: I see, this is handled here. Perhaps you can ignore my previous comment.

modules/acquia_lift_publisher/config/schema/acquia_lift_publisher.schema.yml
```diff
@@ -0,0 +1,7 @@
+acquia_lift_publisher.settings:
+  type: config_object
+  label: 'Acquia Lift Publisher Settings'
+  mapping:
+    personalized_content_only:
+      type: boolean
+      label: 'Push personalizing content only'
```
ynx, line 22: Awesome, thanks!

modules/acquia_lift_publisher/src/EventSubscriber/Publish/PublishOnlyRendered.php
```diff
@@ -0,0 +1,165 @@
...
+use Drupal\acquia_contenthub_publisher\Event\ContentHubEntityEligibilityEvent;
+use Drupal\acquia_lift_publisher\Form\AcquiaLiftPublisherSettingsForm;
+use Drupal\Core\Config\ImmutableConfig;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * Responsible for filtering contents based on publishing configurations.
+ *
+ * @see \Drupal\acquia_lift_publisher\Form\AcquiaLiftPublisherSettingsForm::$pushSettingField
+ * @package Drupal\acquia_lift_publisher\PruneCdf
```
ynx: Is `PruneCdf` intended?

modules/acquia_lift_publisher/src/EventSubscriber/Publish/PublishOnlyRendered.php
c0d39uru: @kaynenh You can delete your comments, if you want.

ynx (changes requested):

ynx:
Great progress, @saxumVermes! You are steamrolling this story solid and fast. ^^

modules/acquia_lift_publisher/src/EventSubscriber/Publish/PublishOnlyRendered.php
```diff
@@ -0,0 +1,165 @@
...
+    $this->publisherSettings = $config;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function getSubscribedEvents() {
+    return [
+      ContentHubPublisherEvents::ENQUEUE_CANDIDATE_ENTITY => ['onEnqueueCandidateEntity', 100],
+      AcquiaContentHubEvents::PRUNE_PUBLISH_CDF_ENTITIES => ['onPrunePublishCdfEntities', 100],
```
ynx: Is this where we set the priority of the event? Let's make sure this event is HIGHLY respected - unless another event is intentionally set to ridiculously high priority, Content Hub modules really want to respect the checkbox of pushing just personalization content.

modules/acquia_lift_publisher/acquia_lift_publisher.module
saxumVermes: Does deletion involve the export queue? It is an automatic process if I'm not mistaken, so it would remove the content from the subscriber site as well, wouldn't it?

modules/acquia_lift_publisher/acquia_lift_publisher.links.task.yml
saxumVermes: When I was planning the form I thought about the possibility that it might be a subject of extension in case any other configuration needs to be added. If you think we should move the checkbox to EntityViewModeConfigurationForm anyway I'll do that instead. Should I, in that case as well? Also as the class suggests, it has to do with entity view configurations, although this is more related to general synchronization process. What do you think?

ynx: I am not expecting any other configuration options at this point. It would be better to introduce the new tab / form only when we know we need. 

I'd recommend that we move the checkbox to EntityViewModeConfigurationForm. I think it can be the last option there. I didn't notice the form name was `EntityViewModeConfigurationForm`. I think it is better to be named `ContentPublishingForm` (or a similar, more precise naming), now.

modules/acquia_lift_publisher/acquia_lift_publisher.module
ynx: I am not so sure. @EclipseGc to weigh in on this one? ^^

kaynenh: I think this is accurate. A delete would automatically remove the entity from Plexus via the acquia_contenthub_publisher module and we are not concerned with the syndication side from a Lift perspective, but if the customer happens to be both a Personalization and Syndication customer, the delete webhook would also delete those entities at the subscriber sites.

modules/acquia_lift_publisher/src/EventSubscriber/Publish/PublishOnlyRendered.php
saxumVermes: Corrected.

modules/acquia_lift_publisher/src/EventSubscriber/Publish/PublishOnlyRendered.php
saxumVermes: Noted.

modules/acquia_lift_publisher/src/EventSubscriber/Publish/PublishOnlyRendered.php
saxumVermes: Noted.

modules/acquia_lift_publisher/src/EventSubscriber/Publish/PublishOnlyRendered.php
saxumVermes: Extended.

modules/acquia_lift_publisher/acquia_lift_publisher.links.task.yml
saxumVermes: Moved and renamed.

modules/acquia_lift_publisher/acquia_lift_publisher.module
saxumVermes: Moved to separate class.

modules/acquia_lift_publisher/acquia_lift_publisher.module
ynx: Right, 8.x-2.x delete is not queued, so we don't have to worry about it here. Resolving this thread.

ynx (approved):

ynx:
Still missing tests but from draft point of view this looks great to me!

modules/acquia_lift_publisher/src/Form/ContentPublishingForm.php
```diff
@@ -198,9 +229,9 @@ protected function setImageSelectorElement(array &$form, $entity_type_id, $bundl
     if (!empty($image_fields)) {
       $form['acquia_lift_preview_image'] = [
         '#type' => 'select',
-        '#title' => t("Select bundle's preview image field."),
+        '#title' => $this->t("Select bundle's preview image field."),
```
Ateee89, line 232: Single quotes please.

saxumVermes: Please note the single quote inside the string.

Ateee89: Missed it :)

kaynenh (approved):
The technical pieces of this look good after looking it over with Kris. We would like the Lift team to own the functional review of this and will be requesting them to do so and merge once they have.

PR #293, opened 2020-02-17
LCH-3782: Lift Publisher should allow exporting of default view modes.

Max-Yudin:


c0d39uru (approved):

modules/acquia_lift_publisher/config/schema/acquia_lift_publisher.entity_config.yml
```diff
@@ -0,0 +1,20 @@
+# Schema for the configuration files of the Acquia Lift Publisher module.
```
kaynenh, line 1: Good catch adding this. Surprised this yml file isn't already in acquia_lift_publisher

modules/acquia_lift_publisher/tests/src/Functional/Form/EntityViewModeConfigurationFormTest.php
```diff
@@ -0,0 +1,102 @@
+<?php
```
kaynenh, line 1: Good test

kaynenh (approved):
This all looks good. Thanks Maks-Yudin!

PR #294, opened 2020-03-17
Add D9 readiness ORCA job

TravisCarden:
This PR is mandatory unless this branch is not intended to survive to Drupal 9--in which, it may be closed unmerged. Thanks.


scor:
8.x-3.x will not be compatible with D9. only 8.x-4.x will be (#295).

scor:
actually, even if we don't support D9 in 3.x we can still bring it up to speed with the latest changes from ORCA. I pulled the relevant parts of your patch in #297. Thanks Travis!

PR #291, opened 2020-01-20
LCH-3698 / LEB-4296: Fix block content rendering.

saxumVermes:

PR #299, opened 2020-05-21
LEB-4401: Only add preview image if it exists

kaynenh:
Only add preview image if it exists, otherwise still export entity without silently failing...

### DIFF
--- a/modules/acquia_lift_publisher/src/EventSubscriber/Cdf/EntityRenderHandler.php
+++ b/modules/acquia_lift_publisher/src/EventSubscriber/Cdf/EntityRenderHandler.php
@@ -180,13 +180,10 @@ class EntityRenderHandler implements EventSubscriberInterface {
           $cdf->addAttribute('view_mode', CDFAttribute::TYPE_STRING, $view_mode);
           $cdf->addAttribute('view_mode_label', CDFAttribute::TYPE_STRING, $display->label());
 
-          if (isset($view_modes['acquia_lift_preview_image'])) {
-            $preview_image = $entity->{$view_modes['acquia_lift_preview_image']}->first();
+          $preview_src = $this->buildPreviewImageAttributeSource($view_modes, 'acquia_lift_preview_image', 'acquia_lift_publisher_preview_image', $entity);
 
-            if ($preview_image) {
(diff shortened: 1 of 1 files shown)


modules/acquia_lift_publisher/src/EventSubscriber/Cdf/EntityRenderHandler.php
```diff
@@ -183,12 +183,10 @@ public function onCreateCdf(CreateCdfEntityEvent $event) {
           if (isset($view_modes['acquia_lift_preview_image'])) {
             $preview_image = $entity->{$view_modes['acquia_lift_preview_image']}->first();
 
-            if (!$preview_image) {
-              continue;
+            if ($preview_image) {
+              $src = ImageStyle::load('acquia_lift_publisher_preview_image')->buildUrl($preview_image->entity->getFileUri());
```
scor: Looks good. I wonder if we should also add another level of conditional to make sure that the imagestyle can actually build a valid URL as opposed to an empty string. Not sure how confident we can be that given an entity, buildUrl() will work.

scor: in other words, not add the `preview_image` attribute to cdf if we have an empty `$src`. Let me know what you think and I can update the PR as well.

kaynenh: Noted. I came up with something a bit different to hopefully protect further.

kaynenh:
Also note, I did not test this so that should probably happen before merging.

scor:
Thanks for all your help @kaynenh - I'll test the code soon.

kaynenh:
Yep! I think the scenario is if you associate a preview image to a bundle but then do not actually add an image to an entity of that bundle, it should still export with this code.

scor:
We will want to wait for #291 to get merged before merging this one as they are both touching the same lines. #291 is pretty large and blocked by some failing tests right now. It should be easy to re-apply the changes here on top of it though once it's merged.

scor:
Moved to #304 in dedicated branch. let's continue the conversation there.

PR #305, opened 2020-06-24
LEB-4601 | update ES type for source_entity, language and view_mode

scor:
Those fields are being renamed to source_entity_id, langcode and view_mode_id respectively. This is to avoid ES index conflicts and to avoid having to reindex
ES.


hzheng06 (approved):
LGTM

scor:
outdated

PR #307, opened 2020-07-14
LEB-3179 | Make Decision API url required and set default to main production endpoint

scor:


baophan1 (approved):
Looks good! Is it necessary to add this as a unit test or is it overkill?

baophan1 (changes requested):
Sorry I realize that if the url is localhost, the url is required to use `http` rather than `https`. We'll need to add an if statement to check for this

hzheng06 (approved):
LGTM!

scor:
> Sorry I realize that if the url is localhost, the url is required to use `http` rather than `https`. We'll need to add an if statement to check for this

This URL is the Decision API endpoint URL, so it should always be https and never localhost.

scor:
> Looks good! Is it necessary to add this as a unit test or is it overkill?

overkill IMO since we are using Drupal core Form API which is well tested in core itself.

baophan1 (approved):
I'm good with the changes if the assumption that localhost will never be used

PR #308, opened 2020-08-14
LCH-4200: Fix configuration translation.

Ateee89:


Ateee89:
One test scenario fails, but cannot find out the reason. Cannot reproduce locally. Will try to figure it out on Monday.

modules/acquia_lift_publisher/tests/src/Kernel/EventSubscriber/Cdf/EntityRenderHandlerTest.php
```diff
@@ -352,7 +354,7 @@ protected function assertCdfAttributes(ContentEntityInterface $entity, array $cd
     foreach ($original_languages as $original_language) {
       $translation = $entity->getTranslation($original_language);
       $orig_label = $translation->label();
-      $this->assertNotFalse(strpos(htmlspecialchars($orig_label), $contents[$original_language]), 'Cdf contains the translated content.');
+      $this->assertNotFalse(strpos(htmlspecialchars($orig_label, ENT_QUOTES), $contents[$original_language]), 'Cdf contains the translated content.');
```
scor: @Ateee89 could you explain why the ENT_QUOTES fixed the problem here? On another branch I had fixed it by reversing the arguments of strpos(). According to the PHP docs is `strpos($haystack, $needle)` so that's why I think it should be reversed ($orig_label is what we're searching inside the CDF $content. but maybe I'm missing something.

Ateee89: @scor When we do not add ENT_QUOTES flag and the random generated string in the label contains a single quote (which wont be converted without the flag) the strpos will return FALSE.
You are right on the reverse part, I did reverse it, although test will fail when the random string generates a label with a single quote. I did test it with a hardcoded label which contains a single quote and it was failing till I added the ENT_QUOTES flag.

scor:
Thanks @Ateee89 , moving to https://github.com/acquia/acquia_lift/pull/309

PR #320, opened 2021-07-12
DIT-295: Update to latest ORCA, test on Drupal 9.2 (Lift 4.x)

aposidelov:


japerry (changes requested):
While I see the code style changes, the other large part of the PR I started was a change to the .travis.yml file to make it compliant with ORCA 3.x standards. I don't see any changes that I made in my 4.x PR in this one. That is critical first before we can look at the other tests, which may not be valid anymore with the new version of ORCA.

aposidelov:
> While I see the code style changes, the other large part of the PR I started was a change to the .travis.yml file to make it compliant with ORCA 3.x standards. I don't see any changes that I made in my 4.x PR in this one. That is critical first before we can look at the other tests, which may not be valid anymore with the new version of ORCA.

Thanks, I've updated .travis.yml how it's implemented in DIT-295

PR #327, opened 2021-11-26
ORCA-203: Bump php version to 7.4 in 8.x-3.x

secretsayan:


TravisCarden (changes requested):
Needs the `distro`.

TravisCarden (approved):

TravisCarden (changes requested):
Oops. I just noticed that the tests are failing because Composer is being downgraded on line 42/45:

```yaml
  - composer self-update 1.10.16
```

That needs to be removed. And I guess we may as well change `ORCA_VERSION` to `^3` while we're at it.

PR #330, opened 2022-03-29
Fixing D9 deprecations.

narendradesai:


japerry (changes requested):
I've merged all the previous PRs around Travis, php, phpcs, etc into this one issue. Its upto @scor or @kk5190 to fix the remaining failing tests.

